PR #17, opened 2022-03-11
Mat 3370

kroos47:
- removed some unwraps where client doesn't panic when a node is stopped or when the chunk size is less than 96


src/main.rs
```diff
@@ -129,9 +129,15 @@ pub async fn main() {
...
-		.expect("failed to get latest block header of chain");
+	let block_header = match rpc::get_chain_header(&cfg.full_node_rpc)
+		.await {
+			Ok(head) => head, 
+			Err(e) => {
+				log::info!("❗❗failed to get latest block header of chain, check if connected to node \n{:?}",e);
+				return
+			},
+		};
+		// .expect("failed to get latest block header of chain, check if connected to node");
```
prabal-banerjee, line 140: Let's remove these comments?

src/main.rs
```diff
@@ -164,7 +170,15 @@ pub async fn main() {
 	let cf_handle_1 = db_3.cf_handle(consts::BLOCK_HEADER_CF).unwrap();
 
 	let read_future = read.for_each(|message| async {
-        let data = message.unwrap().into_data();
+        // let data = message.unwrap().into_data();
```
prabal-banerjee, line 173: Same as above.

src/main.rs
```diff
@@ -180,11 +194,23 @@ pub async fn main() {
                 let max_rows = header.extrinsics_root.rows;
                 let max_cols = header.extrinsics_root.cols;
                 let commitment = header.extrinsics_root.commitment.clone();
-
+				
+				if max_cols<3{
+					log::info!("lower number of chunks");
+					std::process::exit(0);
```
prabal-banerjee, line 200: @MiguelDD1 Should we avoid such exits?

prabal-banerjee (approved):

src/main.rs
```diff
@@ -129,9 +129,15 @@ pub async fn main() {
...
 
-	let block_header = rpc::get_chain_header(&cfg.full_node_rpc)
-		.await
-		.expect("failed to get latest block header of chain");
+	let block_header = match rpc::get_chain_header(&cfg.full_node_rpc)
+		.await {
+			Ok(head) => head, 
+			Err(e) => {
+				log::info!("❗❗failed to get latest block header of chain, check if connected to node \n{:?}",e);
+				return
```
MiguelDD1, line 137: This `return` and the previous `expect` have similar effects at `main` function.
If you want to log any error instead of just panic, my suggestion is to create a inner function and log any error in one single place. Something like:

```Rust
pub async fn main() {
    do_main().await().map_err(|err_msg| log::info!("{}", err_msg));
}

pub async fn do_main() -> Result<(), String> {
  /// ... previous main where `unwrap` was replaced by `?` & `map_err(...)`.
}
```

Having said that, I am **not big fan of using `String` as error type** (in `Result<(), String>`), and **we should use** a better error handling like [thiserror crate](https://crates.io/crates/thiserror) and/or [Anyhow](https://github.com/dtolnay/anyhow).

src/main.rs
MiguelDD1: My suggestion is to use [thiserror create](https://crates.io/crates/thiserror) and/or [Anyhow](https://github.com/dtolnay/anyhow) for error handling.

src/main.rs
luka-ethernal: +1 for anyhow!

src/main.rs
sh3ll3x3c, line 200: Instead of exiting LC, should we keep polling for new blocks for a couple of times before possibly connecting to a new full node? 

This is a wider issue on the LC - how should we proceed when LC encounters errors on full nodes which would trigger potential switching to other (honest or non-malfunctioning) full nodes? cc: @prabal-banerjee

sh3ll3x3c:
@kroos47 we have a .rustfmt config setup, we can use rust fmt for more concise and uniform formatting...

src/main.rs
prabal-banerjee: A particular block data is unavailable (due to no or wrong proof) should make the LC instantly switch to new peers. Even after multiple tries, if it fails to retrieve data, it should report it as a fatal error. WDYT? @sh3ll3x3c

sh3ll3x3c: @prabal-banerjee Fully agree - lets go with a simple retry mechanism with a single full node + switch to a different node if not successful. If if fails a couple of times with other nodes too - fatal error.

Now, the next question would be - do we have a discovery protocol (Libp2p) in place for light clients to discover full nodes (not just other light clients)?

kroos47:
making necessary changes as per commented.

kroos47:
Updated this to [https://github.com/maticnetwork/avail-light/pull/24](PR)

kroos47:
[scscs](url)

kroos47:
Updated this to PR [https://github.com/maticnetwork/avail-light/pull/24](url)

PR #19, opened 2022-03-15
Ava 78 default log level in config

aterentic-ethernal:


prabal-banerjee (approved):
Looks good. Let's create a separate task/PR for saving output to file.

PR #24, opened 2022-03-24
Test/3370

kroos47:
-handing of situation when a node is crashed or if the chunk size is less than 3. 
PS :: the solution for connecting to other nodes as per [https://github.com/maticnetwork/avail-light/pull/17#discussion_r824593668](commented) are not included in this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ libipld = { version = "0.12.0", default-features = false, features = ["dag-cbor"
 multihash = { version = "0.14.0", default-features = false, features = ["blake3"] }
 ipfs-embed = { version = "0.22.4" }
 anyhow = "1.0.41"
+thiserror = "1.0.30"
 tempdir = "0.3.7"
 ed25519-dalek = "1.0.1"
 async-std = { version = "1.9.0", features = ["attributes"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ libipld = { version = "0.12.0", default-features = false, features = ["dag-cbor"
 multihash = { version = "0.14.0", default-features = false, features = ["blake3"] }
 ipfs-embed = { version = "0.22.4" }
 anyhow = "1.0.41"
-thiserror = "1.0.30"
 tempdir = "0.3.7"
 ed25519-dalek = "1.0.1"
 async-std = { version = "1.9.0", features = ["attributes"] }
--- a/src/main.rs
+++ b/src/main.rs
@@ -147,8 +147,8 @@ pub async fn do_main() -> Result<()> {
 	log::info!("Syncing block headers from 0 to {}", latest_block);
 
 	//tokio-tungesnite method for ws connection to substrate.
-	let url = url::Url::parse(&cfg.full_node_ws)?;
-	let (ws_stream, _response) = connect_async(url).await?;
+	let url = url::Url::parse(&cfg.full_node_ws).context("Invalid url: Failed to connect")?;
+	let (ws_stream, _response) = connect_async(url).await.context("Failed to connect to ws")?;
 	let (mut write, mut read) = ws_stream.split();
 
 	// attempt subscription to full node block mining stream
@@ -156,7 +156,7 @@ pub async fn do_main() -> Result<()> {
 		.send(Message::Text(
 			r#"{"id":1, "jsonrpc":"2.0", "method": "subscribe_newHead"}"#.to_string() + "\n",
 		))
-		.await?;
+		.await.context("ws-message(subscribe_newHead) send failed")?;
 
 	let _subscription_result = read.next().await.unwrap().unwrap().into_data();
 	log::info!("Connected to Substrate Node");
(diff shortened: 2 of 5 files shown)


prabal-banerjee:
The above link doesn't work for me. Can you check once? 

Also, I think format check fails in CI. Kindly check.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ libipld = { version = "0.12.0", default-features = false, features = ["dag-cbor"
 multihash = { version = "0.14.0", default-features = false, features = ["blake3"] }
 ipfs-embed = { version = "0.22.4" }
 anyhow = "1.0.41"
+thiserror = "1.0.30"
```
aterentic-ethernal, line 33: Is this used somewhere or we can remove thiserror atm?

src/rpc.rs
```diff
@@ -20,140 +21,100 @@ fn is_secure(url: &str) -> bool {
...
-					Err(_) => None,
-				}
-			} else {
-				let client = hyper::Client::new();
-				match client.request(req).await {
-					Ok(resp) => Some(resp),
-					Err(_) => None,
-				}
-			};
+		.map_err(|e| anyhow::anyhow!("{}", e))?;
```
aterentic-ethernal, line 35: I think we don't need to wrap errors here, instead of 

```
let req = hyper::Request::builder()
		.method(hyper::Method::POST)
		.uri(url)
		.header("Content-Type", "application/json")
		.body(hyper::Body::from(payload))
		.map_err(|e| anyhow::anyhow!("{}", e))?;
```

we should be able to write  just:

```
let req = hyper::Request::builder()
		.method(hyper::Method::POST)
		.uri(url)
		.header("Content-Type", "application/json")
		.body(hyper::Body::from(payload))?;
```

But since we had some context in the error previously, if we import Context trait from anyhow with:

```use anyhow::{Context, Result};```

we should be able to attach context to error also:

```
	let req = hyper::Request::builder()
		.method(hyper::Method::POST)
		.uri(url)
		.header("Content-Type", "application/json")
		.body(hyper::Body::from(payload))
		.context("Failed to build HTTP POST request object")?;
```

If you agree, this should be done on all places instead of creating error with anyhow! macro.

File: Cargo.toml (hunk 1 above)
Comment (kroos47) [reply Cargo.toml:33]: no I will remove it.

File: src/main.rs [hunk 3]
```diff
@@ -239,40 +245,56 @@ pub async fn main() {
...
+				// notify ipfs-based application client
+				// that newly mined block has been received
+				block_tx
+					.send(types::ClientMsg {
+						num,
+						max_rows,
+						max_cols,
+						header,
+					})
+					.unwrap();
```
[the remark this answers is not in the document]
kroos47: okey changing that part also

prabal-banerjee (commented):
It appears to me that the changes reduce the verbosity we had in logs. We should keep (or add) error messages that were present earlier. Not sure whether I misunderstood.

kroos47:
> It appears to me that the changes reduce the verbosity we had in logs. We should keep (or add) error messages that were present earlier. Not sure whether I misunderstood.

Actually added `anyhow::context` for getting some context on error! Let me know if we need to add more to that

src/rpc.rs
kroos47: Done

aterentic-ethernal (approved):

src/main.rs
```diff
@@ -150,80 +147,95 @@ pub async fn main() {
 	log::info!("Syncing block headers from 0 to {}", latest_block);
 
 	//tokio-tungesnite method for ws connection to substrate.
-	let url = url::Url::parse(&cfg.full_node_ws).unwrap();
-	let (ws_stream, _response) = connect_async(url).await.expect("Failed to connect");
+	let url = url::Url::parse(&cfg.full_node_ws)?;
+	let (ws_stream, _response) = connect_async(url).await?;
```
prabal-banerjee, line 151: Add error message here?

src/main.rs
```diff
@@ -150,80 +147,95 @@ pub async fn main() {
...
 	let (mut write, mut read) = ws_stream.split();
 
 	// attempt subscription to full node block mining stream
 	write
 		.send(Message::Text(
 			r#"{"id":1, "jsonrpc":"2.0", "method": "subscribe_newHead"}"#.to_string() + "\n",
 		))
-		.await
-		.unwrap();
+		.await?;
```
prabal-banerjee, line 159: Is there any error message we can show here?

src/main.rs
```diff
@@ -135,9 +134,7 @@ pub async fn main() {
 		log::info!("IPFS backed application client: {}\t{:?}", peer_id, addrs);
 	}
 
-	let block_header = rpc::get_chain_header(&cfg.full_node_rpc)
-		.await
-		.expect("failed to get latest block header of chain");
+	let block_header = rpc::get_chain_header(&cfg.full_node_rpc).await?;
```
prabal-banerjee, line 137: Previously we used to show error: `failed to get latest block header of chain`. Is that still preserved?

prabal-banerjee (approved):

src/main.rs
kroos47: It will show something like `failed to build HTTP POST request object(get_chainHeader)`. Or we can change to old one

src/main.rs
kroos47: Added context here

sh3ll3x3c (approved):

kroos47:
> Small observation: we should discuss and implement retries and exits in a logical and consistent manner at some point in the future.
> 
> As a side note, we should also strive to have smaller PRs.

Yeh I agree.

PR #33, opened 2022-04-15
Ava108/retry

kroos47:
Retry Mechanism for light client

What's needs to be done!

- Exponential Backoff on same node
- Exponential backoff with retry again on another if the connection is lost


kroos47:
https://linear.app/matic/issue/AVA-108/retry-mechanism-for-light-client

src/rpc.rs
```diff
@@ -376,3 +379,51 @@ pub fn get_id_specific_size(num: Block) -> HashMap<u32, u32> {
 	}
 	index
 }
+
+pub async fn check_connection(
+	full_node_ws: Vec<String>,
+) -> Result<Option<WebSocketStream<MaybeTlsStream<TcpStream>>>> {
+	let mut ws_ = None;
+	for x in full_node_ws.iter() {
+		let _url = url::Url::parse(&x)?;
```
aterentic-ethernal, line 388: If we parse URLs before calling this method, error path would be clear and this method would be simpler (instead of result, it would return `Option<WebSocketStream<MaybeTlsStream<TcpStream>>>`

src/main.rs
```diff
@@ -129,174 +129,195 @@ pub async fn do_main() -> Result<()> {
...
 		log::info!("IPFS backed application client: {}\t{:?}", peer_id, addrs);
-	}
-
-	let block_header = rpc::get_chain_header(&cfg.full_node_rpc).await?;
+	};
+	let rpc_url = match rpc::check_http(cfg.full_node_rpc).await {
+		Ok(a) => a,
+		Err(e) => return Err(e),
+	};
+	let rpc_: &str = &*rpc_url;
```
aterentic-ethernal, line 139: Is it possible to reduce this to something like:

`let rpc_url = match rpc::check_http(cfg.full_node_rpc).await?.clone();`?

aterentic-ethernal, line 139: Is it possible to reduce this to something like:

`let rpc_url = match rpc::check_http(cfg.full_node_rpc).await?.clone();`?

kroos47:
New one with rebasing

PR #34, opened 2022-04-17
Parse extrinsic and commitment as hex-encoded strings instead of arrays.

luka-ethernal:


sh3ll3x3c (approved):

aterentic-ethernal (approved):
There is one unused import warning from github actions, otherwise 👍

PR #39, opened 2022-04-26
add: HTTP for data query

kroos47:
- http for getting whole data from full node via `curl -s localhost:9000/v1/data/_blocknumber`
- http for getting app related data from full node via `curl -s localhost:3001/v1/appdata/_blocknumber`

Not connected to IPFS for now and using the full node rpc to query the data.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,0 +1,151 @@
+extern crate rocksdb;
+
+use std::{
+	pin::Pin,
+	sync::Arc,
+	task::{Context, Poll},
+};
+
+use ::futures::prelude::*;
+use chrono::{DateTime, Local};
+use hyper::{
+	header::ACCESS_CONTROL_ALLOW_ORIGIN, service::Service, Body, Method, Request, Response, Server,
+	StatusCode,
+};
+use regex::Regex;
+use rocksdb::{DBWithThreadMode, SingleThreaded};
+use tokio;
+
+use crate::rpc::{check_http, get_block_by_number, get_kate_query_proof_by_cell};
+
+// 💡 HTTP part where handles the RPC Queries
+
+//service part of hyper
+struct Handler {
+	store: Arc<DBWithThreadMode<SingleThreaded>>,
+	url: String,
+}
+
+impl Service<Request<Body>> for Handler {
+	type Error = hyper::Error;
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+	type Response = Response<Body>;
+
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
+	}
+
+	fn call(&mut self, req: Request<Body>) -> Self::Future {
+		fn match_url_block(path: &str) -> Result<u64, String> {
+			let re = Regex::new(r"^(/v1/data/(\d{1,}))$").unwrap();
+			if let Some(caps) = re.captures(path) {
+				if let Some(block) = caps.get(2) {
+					return Ok(block.as_str().parse::<u64>().unwrap());
+				}
+			}
+			Err("no match found !".to_owned())
+		}
+
+		fn mk_response(s: String) -> Result<Response<Body>, hyper::Error> {
+			Ok(Response::builder()
+				.status(200)
+				.header(ACCESS_CONTROL_ALLOW_ORIGIN, "*")
+				.header("Content-Type", "application/json")
+				.body(Body::from(s))
+				.unwrap())
+		}
+
+		let local_tm: DateTime<Local> = Local::now();
+		log::info!(
+			"⚡️ {} | {} | {}",
+			local_tm.to_rfc2822(),
+			req.method(),
+			req.uri().path()
+		);
+
+		let _db = self.store.clone();
+		let url = self.url.clone();
+
+		Box::pin(async move {
+			let res = match req.method() {
+				&Method::GET => {
+					if let Ok(block_num) = match_url_block(req.uri().path()) {
+						//cell logic to be written here
+						let block = get_block_by_number(&url, block_num).await.unwrap();
+						let header = block.header;
+						let max_rows = header.extrinsics_root.rows;
+						let max_cols = header.extrinsics_root.cols;
+						let mut vec = Vec::new();
+						for i in 0..max_rows {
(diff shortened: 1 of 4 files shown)


src/types.rs
```diff
@@ -330,12 +332,13 @@ impl Default for RuntimeConfig {
 		RuntimeConfig {
 			http_server_host: "127.0.0.1".to_owned(),
 			http_server_port: 7000,
+			http_data_port: 9000,
```
prabal-banerjee, line 335: Why do we need different ports? Can't we serve on the same port (7000) by mapping to different URLs?

src/api.rs
```diff
@@ -0,0 +1,151 @@
...
+					if let Ok(block_num) = match_url_block(req.uri().path()) {
+						//cell logic to be written here
+						let block = get_block_by_number(&url, block_num).await.unwrap();
+						let header = block.header;
+						let max_rows = header.extrinsics_root.rows;
+						let max_cols = header.extrinsics_root.cols;
+						let mut vec = Vec::new();
+						for i in 0..max_rows {
+							for j in 0..max_cols {
+								let cells = get_kate_query_proof_by_cell(&url, block_num, i, j)
```
prabal-banerjee, line 81: Looks like we are re-fetching all the relevant cells without relying on existing code which already fetches the data. Or am I missing something?

prabal-banerjee (changes requested):

src/main.rs
```diff
@@ -18,6 +18,8 @@ use tokio_tungstenite::tungstenite::protocol::Message;
 
 use crate::http::calculate_confidence;
 
+mod api;
+mod api_2;
```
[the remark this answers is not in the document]
kroos47: okey

src/types.rs
kroos47: Well I tried but it was not working for some reason! So I just created new ports. I will dig deeper into this

src/api.rs
kroos47: I tried to fetch the HEADER from the DB that stores the headers and the confidence. But the problem was that I was not able to retrieve that properly like we want and I was also rushing for the product team. Anyways I will try again to fetch from the DB

src/api_2.rs
```diff
@@ -0,0 +1,207 @@
...
+			req.uri().path()
+		);
+
+		let _db = self.store.clone();
+		let url = self.url.clone();
+
+		Box::pin(async move {
+			let res = match req.method() {
+				&Method::GET => {
+					if let Ok(block_num) = match_url_block(req.uri().path()) {
```
sh3ll3x3c, line 113: We should try to avoid having nested matches as much as possible. What we can do in those cases is use Option::and_then.

References:

https://doc.rust-lang.org/rust-by-example/error/option_unwrap/and_then.html
https://stackoverflow.com/a/71269133

src/api.rs
sh3ll3x3c: Yeah, we should def have an abstraction over the fetching logic, as it will be used in at least a couple of places...

src/api.rs
```diff
@@ -0,0 +1,151 @@
...
+						.insert(ACCESS_CONTROL_ALLOW_ORIGIN, "*".parse().unwrap());
+					Ok(not_found)
+				},
+			};
+			res
+		})
+	}
+}
+
+struct MakeHandler {
```
sh3ll3x3c, line 113: We should probably have a more descriptive struct name here. What does it do, what does it handle? That should help in making more readable names in code...

src/api_2.rs
```diff
@@ -0,0 +1,207 @@
...
+	type Error = hyper::Error;
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+	type Response = Response<Body>;
+
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
+		Poll::Ready(Ok(()))
+	}
+
+	fn call(&mut self, req: Request<Body>) -> Self::Future {
+		fn match_url_block(path: &str) -> Result<u64, String> {
```
sh3ll3x3c, line 44: Should we extract these functions and have unit tests for them externally? 

Do we need unwraps here or some more appropriate error handling?

src/api_2.rs
```diff
@@ -0,0 +1,207 @@
...
+	store: Arc<DBWithThreadMode<SingleThreaded>>,
+	url: String,
+}
+
+impl<T> Service<T> for MakeHandler {
+	type Error = hyper::Error;
+	type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+	type Response = Handler;
+
+	fn poll_ready(&mut self, _: &mut Context) -> Poll<Result<(), Self::Error>> {
```
sh3ll3x3c, line 180: Unsure if we really need this function, we could just call the `Poll::Ready` directly.

src/api.rs
kroos47: Fetching from rocksdb was working, But we were using `get_kate_proof` which fetches the relevent cells but, in case of no app_id is given it fetches the random cells, which is not the data that we want. That's what I thought on it

src/api_2.rs
kroos47: well fn call() is needed which used by the hyper [crate](https://docs.rs/hyper/latest/hyper/service/trait.Service.html). Atleast thats my understanding

src/api_2.rs
kroos47: same in the case if poll_ready()

src/api_2.rs
sh3ll3x3c: @kroos47 ofc, but talking about the `match_url_block` func...

kroos47: oh okey. we can do that 👍

kroos47:
> * http for getting whole data from full node via `curl -s localhost:9000/v1/data/_blocknumber`
> * http for getting app related data from full node via `curl -s localhost:3001/v1/appdata/_blocknumber`
> 
> Not connected to IPFS for now and using the full node rpc to query the data.

Update::
- Using the DB for fetching blocks
- `curl -s localhost:9000/v1/appdata/_blocknumber_/_appid_` for id related data and 
   `curl -s localhost:9000/v1/data/_blocknumber_/-1` for full data on a block
   `curl -s localhost:9000/v1/confidence/_blocknumber_ `for confidence

PR #40, opened 2022-05-02
Miguel/ava 154 proof verification fails on light

MiguelDD1:
- [x] Dependencies:
  - [x] Support TLS for WSS
  - [x] `kate-proof` & `kate-recovery` selected from specific branch.
  - [x] Compilation issue on `ipfs_embed:0.22.4`
  - [x] Use Multithread RockDB.
- [x] Listen on finalized headers.
- [x] Use `?` and new `error::App` to simplify error handling.
- [x] `main.rs` simplified and code dedup.
- [x] Remove clippy warnings.


prabal-banerjee (approved):

itzmeanjan (approved):
Looks good :+1:

kroos47 (approved):

Cargo.toml
```diff
@@ -7,9 +7,9 @@ authors = ["Polygon Avail Team"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-kate-recovery = {git = "https://github.com/maticnetwork/avail.git"}
-kate-proof = {git = "https://github.com/maticnetwork/avail.git"}
-tokio-tungstenite = "0.16.1"
+kate-recovery = {git = "https://github.com/maticnetwork/avail.git", branch = "miguel/ava-154-proof-verification-fails-on-light"}
+kate-proof = {git = "https://github.com/maticnetwork/avail.git", branch = "miguel/ava-154-proof-verification-fails-on-light"}
```
aterentic-ethernal, line 11: Maybe it makes sense to introduce versioning and tag avail with v0.1.0 instead of using long lived branches?

src/client.rs
```diff
@@ -395,162 +318,119 @@ pub async fn run_client(
...
-			Err(e) => {
-				log::info!("Error encountered while listening for blocks: {}", e);
-				break;
-			},
-		};
+		let rpc_ = check_http(&full_node_rpc).await?;
+		let block_cid_entry = get_block_cid_entry(store, block.num as i128).map(|pair| pair.cid);
+
+		let ipfs_cells = get_matrix(ipfs, block_cid_entry)
+			.inspect_err(|err| log::error!("Fail to fetch cells from IPFS: {}", err))
```
aterentic-ethernal, line 333: We don't make distinction between BlockNotFound and other error causes, which is maybe fine, but it can raise false alarms if log level is error. We should handle it as specific case or just set it to info? Or just put TODO and handle it later on all places where ipfs.get is called?

src/client.rs
```diff
@@ -65,326 +63,251 @@ pub async fn run_client(
...
-
-						// received message was not decodable !
-					},
-				},
-				None => {
-					break;
-				},
-			}
-		}
+		fact_topic.for_each(|msg| {
```
aterentic-ethernal, line 85: Not sure why we break out of a loop if next is None, and why it could be none, but just to be sure, does for_each keeps same meaning?

src/http.rs
```diff
@@ -32,8 +33,81 @@ pub fn serialised_confidence(block: u64, factor: f64) -> String {
 // 💡 HTTP part where handles the RPC Queries
 
 //service part of hyper
+#[derive(From)]
 struct Handler {
-	store: Arc<DBWithThreadMode<SingleThreaded>>,
+	store: Arc<DB>,
+}
+
+impl Handler {
```
sh3ll3x3c, line 41: Isn't the http logic already under development by @kroos47 ?

src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+	HttpServer(#[from] HttpServer),
+	#[error("IPFS error: {0}")]
+	Ipfs(#[from] IpfsClient),
+
+	#[error("Cell (r:{0},c:{1}) data is not available")]
+	CellDataNotAvailable(u16, u16),
+}
+
+#[inline]
+pub fn into_warning<E: Display>(e: &E) { log::warn!("{}", e) }
```
sh3ll3x3c, line 109: Do we really need this level of abstraction here?

src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[inline]
+pub fn into_error<E: Display>(e: &E) { log::error!("{}", e) }
+
+/// Return Err of the expression: `return Err($expression);`.
+///
+/// Used as `fail!(expression)`.
+#[macro_export]
+macro_rules! fail {
```
sh3ll3x3c, line 118: Same as above, are we unnecessarily decreasing the code readability here?

src/http.rs
kroos47: just a code refactoring is okey I guess. Not sure on merge conflicts 😅

PR #45, opened 2022-05-11
Ava 168

kroos47:
- [x] fix on random sampling

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -184,8 +184,7 @@ pub async fn do_main() -> Result<()> {
 					let index_tuple = header.app_data_lookup.index.clone();
 					let kate_cells = rpc::generate_random_cells(max_rows, max_cols, num);
 					//hyper request for getting the kate query request
-					let cells =
-						rpc::get_kate_proof(&rpc_url, num, max_rows, max_cols, kate_cells).await?;
+					let cells = rpc::get_kate_proof(&rpc_url, num, kate_cells).await?;
 					//hyper request for verifying the proof
 					let count = proof::verify_proof(
 						num,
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -184,9 +184,10 @@ pub async fn do_main() -> Result<()> {
 						log::error!("chunk size less than 3");
 					}
 					let commitment = header.extrinsics_root.commitment.clone();
+					let index_tuple = header.app_data_lookup.index.clone();
+					let kate_cells = rpc::generate_random_cells(max_rows, max_cols, num);
```
sh3ll3x3c, line 188: great change!

src/rpc.rs
```diff
@@ -333,30 +312,30 @@ pub fn generate_app_specific_cells(
...
 			row: rows as u16,
 			col: cols as u16,
 			..Default::default()
 		});
-	}
+	});
 	buf
 }
 
-pub fn get_id_specific_size(num: Block) -> HashMap<u32, u32> {
```
aterentic-ethernal, line 357: Maybe to rename num to header or something like that. I think that `num` has confused me at some point somewhere in a codebase :D

kroos47: yeh me too, just went with the flow. Will change

src/main.rs
```diff
@@ -212,31 +213,54 @@ pub async fn do_main() -> Result<()> {
...
+									for j in 0..max_cols {
+										cells.push(types::Cell {
+											block: num,
+											row: i,
+											col: j,
+											..Default::default()
+										});
+									}
+								}
+								cells
```
aterentic-ethernal, line 231: Maybe shorter version (not sure which one is better):

`(0..cols).flat_map(move |col| (0..rows).map(move |row| types::Cell { block: num, row, col, ..Default::default() });` 

Also we could move this to typed mod?

aterentic-ethernal (approved):

src/main.rs
kroos47: yeh thanks! I will try it out

kroos47:
There is this case in `-1` that the number of cells that can be queried must be set or 50% of the total number of cells so that it can be reconstructed later.

PR #50, opened 2022-05-13
fix:kate_query limited verification

kroos47:
- [x] fix on kate_query rpc for cases more than 30 cells and verification on the same cells

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -182,7 +182,6 @@ pub async fn do_main() -> Result<()> {
 					}
 					let commitment = header.extrinsics_root.commitment.clone();
 					let index_tuple = header.app_data_lookup.index.clone();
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -259,9 +259,31 @@ pub async fn do_main() -> Result<()> {
 								vec![]
 							},
 						};
-						let cells = rpc::get_kate_proof(&rpc_url, num, query_cells).await?;
-						let _count =
-							proof::verify_proof(num, max_rows, max_cols, cells, commitment.clone());
+						if query_cells.len() < 30 {
```
prabal-banerjee, line 262: can't we merge the two conditions and let the `for` statement handle it?

luka-ethernal (approved):

PR #61, opened 2022-06-01
Verify proofs on app client.

aterentic-ethernal:


src/sync.rs
```diff
@@ -149,6 +153,31 @@ pub async fn sync_block_headers(
 							count.to_be_bytes(),
 						)
 						.unwrap();
+
+					// Push the randomly selected cells to IPFS
```
sh3ll3x3c, line 157: Are we pushing just the RPC retrieved cells or all of them?

src/app_client.rs
```diff
@@ -51,10 +56,55 @@ async fn process_block(
...
+			log::info!(
+				"Completed {} verification rounds for block {}",
+				count,
+				block.number
+			);
+
+			// TODO: If there are some invalid cells should we fail?
+
+			for cell in cells {
+				if let Err(error) = ipfs.insert(cell.clone().to_ipfs_block()) {
```
sh3ll3x3c, line 86: Same as with sync, we should make sure that we're sending just the RPC retrieved cells, as not to spam the network with duplicates...

src/app_client.rs
```diff
@@ -51,10 +56,55 @@ async fn process_block(
...
+			cells.extend(rpc_fetched);
+
+			if positions.len() > cells.len() {
+				return Err(anyhow!(
+					"Failed to fetch {} cells",
+					positions.len() - cells.len()
+				));
+			}
+
+			let count = verify_proof(
```
sh3ll3x3c, line 70: @prabal-banerjee for cells belonging to the same appID (sequentially ordered), does it make sense at all to verify proof, or should the app client just check the block confidence and then proceed with the reconstruction?

src/app_client.rs
aterentic-ethernal: Sure Ill change that, on both places.

src/app_client.rs
sh3ll3x3c: on the other hand, it does make sense to verify the proof, but the sequential cell proof verification mustn't be used for confidence score. So we should be good here...

src/app_client.rs
aterentic-ethernal: On three places actually.

sh3ll3x3c (approved):

src/app_client.rs
prabal-banerjee: IMO, the app clients should verify all downloaded cells. The confidence score should be just computed on the LC and not the app client.

PR #69, opened 2022-06-13
Fix multiplication with overflow issue on matrix of maximum size.

aterentic-ethernal:
By returning maximum block dimensions instead of calculating next power of two, and padding block to maximum size with random numbers, I was able to cause panics on multiplication overflow in three cases. Cases in get_cells are not reproduced due to time needed to build 130K proofs on avail (it seems that more than 1 hour is needed).

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -21,7 +21,8 @@ fn construct_cell(
 	row_count: u16,
 	cells: Arc<Vec<Option<Vec<u8>>>>,
 ) -> Result<BaseCell, String> {
-	cells[(col * row_count + row) as usize]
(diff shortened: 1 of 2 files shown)


luka-ethernal (approved):

aterentic-ethernal:
> Proof gen for full matrix in release mode takes ~1 hour?

Yes, from logs we can assume that proof generation for one cell takes around 5ms, but since we have 130K rpc calls + proof generation, for fat client and full matrix size, it takes at least 45 minutes. And after that, proof verification of 130K cells is started on LC 🤦‍♂️

PR #80, opened 2022-07-01
Random Sampling change - AVA-248

kroos47:


src/main.rs
```diff
@@ -181,11 +181,13 @@ pub async fn do_main() -> Result<()> {
 		db.clone(),
 		ipfs.clone(),
 		cfg.max_parallel_fetch_tasks,
+		cfg.confidence,
```
prabal-banerjee, line 184: Can we put some checks on the input value? For example, if someone supplies a negative confidence. Not necessarily inside `main`.

src/rpc.rs
```diff
@@ -121,13 +121,13 @@ pub async fn get_block_by_number(url: &str, block: u64) -> Result<Block> {
 	}
 }
 
-pub fn generate_random_cells(max_rows: u16, max_cols: u16) -> Vec<Position> {
+pub fn generate_random_cells(max_rows: u16, max_cols: u16, cell_count: u32) -> Vec<Position> {
 	let max_cells = (max_rows as u32) * (max_cols as u32);
-	let count: u16 = if max_cells < 8 {
+	let count: u16 = if max_cells < cell_count as u32 {
 		// Multiplication cannot overflow since result is less than 8
 		max_rows * max_cols
```
prabal-banerjee, line 128: Can we print a `debug` log when this case happens so that it is easier to debug later?

src/sync_client.rs
```diff
@@ -65,7 +66,8 @@ async fn process_block(
 	let commitment = block_body.header.extrinsics_root.commitment;
 	let block_num = block_body.header.number;
 	// now this is in `u64`
-	let positions = rpc::generate_random_cells(max_rows, max_cols);
+	let cell_count = -((1f64 - confidence / 100f64).log2()).ceil() as u32;
```
prabal-banerjee, line 69: The formula appears in two places. For better maintainability, can we put it into a separate function and call from two places?

src/light_client.rs
```diff
@@ -57,7 +58,13 @@ pub async fn run(
 					}
 					let commitment = header.extrinsics_root.commitment.clone();
 
-					let positions = rpc::generate_random_cells(max_rows, max_cols);
+					let cell_count: u32;
+					if confidence == 100f64 {
+						cell_count = -((1f64 - 99.9f64 / 100f64).log2()).ceil() as u32;
```
prabal-banerjee, line 63: Should we increase precision in this case? Or do we want a conservative default?

src/light_client.rs
```diff
@@ -57,7 +58,13 @@ pub async fn run(
...
+					} else {
+						cell_count = (-((1f64 - confidence / 100f64).log2())).ceil() as u32;
```
prabal-banerjee, line 65: Can we write test cases to ascertain correctness? For example, giving `confidence=60` may give output `1` whereas `1` sample can only give atmost `50%` confidence.

prabal-banerjee (changes requested):

src/light_client.rs
kroos47: yeh sure! But how its being checked is if we put 60 as confidence then it checks for number on cells which gives >=60.

src/light_client.rs
```diff
@@ -57,7 +58,13 @@ pub async fn run(
...
+					}
```
aterentic-ethernal, line 66: Can we also move code to separate function, e.g. `cell_count_for_confidence` or something better?

aterentic-ethernal: Actually, already mentioned :D

src/light_client.rs
prabal-banerjee: I don't think `cell_count` of `1` gives `>=60` confidence.

src/rpc.rs
```diff
@@ -121,13 +121,18 @@ pub async fn get_block_by_number(url: &str, block: u64) -> Result<Block> {
 	}
 }
 
-pub fn generate_random_cells(max_rows: u16, max_cols: u16) -> Vec<Position> {
+pub fn generate_random_cells(max_rows: u16, max_cols: u16, cell_count: u32) -> Vec<Position> {
 	let max_cells = (max_rows as u32) * (max_cols as u32);
-	let count: u16 = if max_cells < 8 {
+	let count: u16 = if max_cells < cell_count as u32 {
 		// Multiplication cannot overflow since result is less than 8
```
sh3ll3x3c, line 127: Should we remove this comment (not relevant anymore)?

src/rpc.rs
```diff
@@ -121,13 +121,18 @@ pub async fn get_block_by_number(url: &str, block: u64) -> Result<Block> {
 	}
 }
 
-pub fn generate_random_cells(max_rows: u16, max_cols: u16) -> Vec<Position> {
+pub fn generate_random_cells(max_rows: u16, max_cols: u16, cell_count: u32) -> Vec<Position> {
```
sh3ll3x3c, line 124: Do we need some sort of sanity check for the cell count, ie in order not to overflow the multiplication?

kroos47: yeh this needs to be done!!

kroos47:
opening a rebased one

PR #85, opened 2022-07-06
Ava 247

kroos47:


src/proof.rs
```diff
@@ -14,22 +15,16 @@ fn kc_verify_proof_wrapper(
...
-		Ok(verification) => {
-			let public_params_hash =
-				hex::encode(sp_core::blake2_128(verification.public_params.as_slice()));
-			let public_params_len = hex::encode(verification.public_params.as_slice()).len();
+	match kate_proof::kc_verify_proof(col as u32, proof, commitment, total_rows, total_cols, &pp) {
+		Ok(_) => {
+			let raw_pp = pp.to_raw_var_bytes();
+			let public_params_hash = hex::encode(sp_core::blake2_128(&raw_pp));
+			let public_params_len = hex::encode(raw_pp).len();
 			log::trace!("Public params ({public_params_len}): hash: {public_params_hash}");
```
aterentic-ethernal, line 25: Maybe it makes sense to move logging of public params to main, since by design, these logs will always be the same?

src/proof.rs
```diff
@@ -14,22 +15,16 @@ fn kc_verify_proof_wrapper(
...
-				Ok(()) => {
-					log::trace!("Verified cell ({row}, {col}) of block {block_num}");
-				},
-				Err(verification_err) => {
-					log::error!("Verification for cell ({row}, {col}) of block {block_num} failed: {verification_err}");
-				},
-			}
-			verification.status.is_ok()
+			log::trace!("Verified cell ({row}, {col}) of block {block_num}");
+			true
```
aterentic-ethernal, line 27: We have to return verification status, since verification can fail (from matched Ok(verification)).

src/proof.rs
kroos47: yeh so that also makes passing params easier!

aterentic-ethernal (approved):
Just a reminder to update avail revision in Cargo.toml once public params related changes are squash merged into avail's develop branch.

PR #99, opened 2022-07-28
Ava 279/http

kroos47:
- /v1/latest_block 
- Incase the block is not synced it returns Not Finalised message

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -5,7 +5,6 @@ use std::{
 };
 
 use anyhow::{Context, Result};
-// use async_std::channel::Receiver;
 use codec::Decode;
 use num::{BigUint, FromPrimitive};
 use rocksdb::DB;
@@ -54,14 +53,11 @@ where
 	Normal(T),
 	NotFound,
 	NotFinalized,
+	InProcess,
 	Error(anyhow::Error),
 }
 
-fn confidence(
-	block_num: u64,
-	db: Arc<DB>,
-	counter: Arc<Mutex<u64>>,
-) -> ClientResponse<ConfidenceResponse> {
+fn confidence(block_num: u64, db: Arc<DB>, counter: u64) -> ClientResponse<ConfidenceResponse> {
 	info!("Got request for confidence for block {block_num}");
 	let res = match get_confidence_from_db(db, block_num) {
 		Ok(Some(count)) => {
@@ -74,8 +70,8 @@ fn confidence(
 			})
 		},
 		Ok(None) => {
-			let lock = counter.lock().unwrap();
-			if block_num < *lock {
+			// let lock = counter.lock().unwrap();
+			if block_num < counter {
 				return ClientResponse::NotFinalized;
 			} else {
 				return ClientResponse::NotFound;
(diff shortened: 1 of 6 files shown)


src/http.rs
```diff
@@ -1,10 +1,11 @@
 use std::{
 	net::SocketAddr,
 	str::FromStr,
-	sync::{mpsc::SyncSender, Arc},
+	sync::{mpsc::SyncSender, Arc, Mutex},
 };
 
 use anyhow::{Context, Result};
+// use async_std::channel::Receiver;
```
aterentic-ethernal, line 8: Not needed I guess.

src/http.rs
```diff
@@ -47,10 +53,15 @@ where
...
 	NotFound,
+	NotFinalized,
 	Error(anyhow::Error),
 }
 
-fn confidence(block_num: u64, db: Arc<DB>) -> ClientResponse<ConfidenceResponse> {
+fn confidence(
+	block_num: u64,
+	db: Arc<DB>,
+	counter: Arc<Mutex<u64>>,
```
aterentic-ethernal, line 63: Is there an option to pass u64 value to confidence and latest_block methods, and read counter before fn call?

src/light_client.rs
```diff
@@ -156,6 +157,8 @@ pub async fn run(
 					block_tx
 						.send(types::ClientMsg::from(params.header))
 						.context("Failed to send block message")?;
+					let mut lock = counter.lock().unwrap();
```
aterentic-ethernal, line 160: Is unwrap OK or we can go with error here?

src/http.rs
```diff
@@ -62,18 +73,41 @@ fn confidence(block_num: u64, db: Arc<DB>) -> ClientResponse<ConfidenceResponse>
...
+	res
+}
+
 fn appdata(
 	block_num: u64,
 	db: Arc<DB>,
 	cfg: &RuntimeConfig,
+	counter: Arc<Mutex<u64>>,
 ) -> ClientResponse<ExtrinsicsDataResponse> {
+	let lock = counter.lock().unwrap();
```
sh3ll3x3c, line 107: Should we use `unwrap` here? Or some other mean of error handling? Don't know what are the consequences of not being able to acquire the lock here...

src/http.rs
```diff
@@ -106,7 +140,14 @@ fn appdata(
 			extrinsics: data,
 		}),
 
-		Ok(None) => ClientResponse::NotFound,
+		Ok(None) => {
+			let lock = counter.lock().unwrap();
```
sh3ll3x3c, line 144: Same...

src/http.rs
```diff
@@ -137,22 +183,32 @@ pub async fn run_server(
...
 	_cell_query_tx: SyncSender<CellContentQueryPayload>,
+	counter: Arc<Mutex<u64>>,
 ) {
 	let host = cfg.http_server_host.clone();
 	let port = cfg.http_server_port;
 
 	let get_mode =
 		warp::path!("v1" / "mode").map(move || warp::reply::json(&Mode::from(cfg.app_id)));
+	let counter_clone = counter.clone();
+	let counter_clone1 = counter.clone();
```
sh3ll3x3c, line 194: Do we need both of these clones, or can we just clone in place the `counter` var when passing it to the func?

src/http.rs
kroos47: changing the function to take u64 instead of Arc<>

src/http.rs
kroos47: as the values are moving in functions, I don't think without this would work! Still might be wrong

aterentic-ethernal (approved):
All cool, I would just rename clones to more descriptive names and that all.

sh3ll3x3c (approved):

PR #118, opened 2022-08-24
randomize ipfs seed

kroos47:
ipfs_seed is not set => random seed is generated

ipfs_seed is set to "random" => random seed is generated

ipfs_seed is set => seed read from configuration is used


config1.yaml
```diff
@@ -0,0 +1,16 @@
+http_server_host = '127.0.0.1'
```
sh3ll3x3c, line 1: Should we commit this (config) file?

src/main.rs
```diff
@@ -171,15 +171,13 @@ pub async fn do_main() -> Result<()> {
...
-
-	let seed = if cfg.ipfs_seed_randomize {
-		let seed: u64 = thread_rng().gen();
-		info!("Using random seed: {seed}");
-		seed
-	} else {
-		cfg.ipfs_seed
+	let seed = match cfg.ipfs_seed.clone() {
+		Some(seed) => {
+			let x = rpc::integer_part(&seed).unwrap();
```
sh3ll3x3c, line 176: Do we need this temp variable?

src/main.rs
```diff
@@ -171,15 +171,13 @@ pub async fn do_main() -> Result<()> {
...
 		cfg.ipfs_port.0
 	};
-
-	let seed = if cfg.ipfs_seed_randomize {
-		let seed: u64 = thread_rng().gen();
-		info!("Using random seed: {seed}");
-		seed
-	} else {
-		cfg.ipfs_seed
+	let seed = match cfg.ipfs_seed.clone() {
```
sh3ll3x3c, line 174: Do we have the case where the seed is set to 'random'? Do we even need it?

src/types.rs
```diff
@@ -474,8 +473,7 @@ pub struct RuntimeConfig {
 	#[serde(default)]
 	#[serde(with = "port_range_format")]
 	pub http_server_port: (u16, u16),
-	pub ipfs_seed_randomize: bool,
-	pub ipfs_seed: u64,
+	pub ipfs_seed: Option<String>,
```
sh3ll3x3c, line 476: If we decide for the seed not to have 'random' option, we could set this var to be Option<u64>, in which case we wouldn't need the `integer_part` func.

src/main.rs
sh3ll3x3c: Also, should we use unwrap here at all?

config1.yaml
kroos47: my bad! removing this 😅

src/main.rs
kroos47: well AFAIK the random is used strictly for docker and that was the reason why I switched to Option<u64> to string

src/main.rs
kroos47: yeh removing that also!

src/main.rs
aterentic-ethernal: I'm not sure if we can go with some "0" value to avoid having different type for seed (instead of "random"). Do we need to discuss this?

src/types.rs
```diff
@@ -474,8 +473,7 @@ pub struct RuntimeConfig {
 	#[serde(default)]
 	#[serde(with = "port_range_format")]
 	pub http_server_port: (u16, u16),
-	pub ipfs_seed_randomize: bool,
-	pub ipfs_seed: u64,
+	pub ipfs_seed: Option<u64>,
```
aterentic-ethernal, line 476: Instead of comment below, we could put `///` comment here, which will be available as `cargo doc`:

E.g.
```
	/// Seed for IPFS keypair. If not set, or seed is 0, random seed is generated
	pub ipfs_seed: Option<u64>,
```

src/types.rs
```diff
@@ -568,8 +566,8 @@ impl Default for RuntimeConfig {
 			http_server_host: "127.0.0.1".to_owned(),
 			http_server_port: (7000, 0),
 			ipfs_port: (37000, 0),
-			ipfs_seed: thread_rng().gen(),
-			ipfs_seed_randomize: false,
+			//0  is used to random configuration of seed
```
aterentic-ethernal, line 569: I guess we can remove this comment if we comment structure itself.

src/main.rs
```diff
@@ -171,15 +171,20 @@ pub async fn do_main() -> Result<()> {
...
+			if seed == 0 {
+				info!("Using random seed");
+				thread_rng().gen()
+			} else {
+				info!("Using seed: {seed}");
+				seed
+			}
+		},
+		None => thread_rng().gen(),
 	};
```
aterentic-ethernal, line 186: Maybe more compact version:

```
	let seed = match cfg.ipfs_seed {
		None => thread_rng().gen(),
		Some(0) => thread_rng().gen(),
		Some(value) => value,
	};
```

src/main.rs
```diff
@@ -171,15 +171,20 @@ pub async fn do_main() -> Result<()> {
...
-
+	println!("💁 Seed: {}\n", seed);
```
aterentic-ethernal, line 187: Maybe just turn this to info! (if we need to log the seed)?

aterentic-ethernal (approved):

PR #135, opened 2022-09-21
Update version and avail dependencies.

aterentic-ethernal:


Cargo.toml
```diff
@@ -1,14 +1,14 @@
...
+version = "2.0.0"
 edition = "2021"
 authors = ["Polygon Avail Team"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-kate-recovery = { git = "https://github.com/maticnetwork/avail.git", rev = "76e2b45" }
-kate-proof = { git = "https://github.com/maticnetwork/avail.git", rev = "76e2b45" }
+kate-recovery = { git = "https://github.com/maticnetwork/avail.git", rev = "062d946" }
```
prabal-banerjee, line 10: Can we use `tag` instead of `rev`? Looks a bit cleaner IMO.

Cargo.toml
```diff
@@ -1,14 +1,14 @@
 [package]
 name = "avail-light"
-version = "0.1.0"
+version = "2.0.0"
```
prabal-banerjee, line 3: The prev version was `0.1`. Avail version now is `1.1`. Should we bump this to `1.0` instead of `2.0`?

kroos47:
Should we cherrypick the commit from develop to support data_root w.r.t the merge happened in avail:main?

Cargo.toml
sh3ll3x3c: @prabal-banerjee the old tag was never actually updated to 1.0 before. As we discussed, the amount of changes we introduced with the new architecture is such that this is more of a v2.0 than 1.1, so it might be better to bump it to 2.0 here.

sh3ll3x3c (approved):

prabal-banerjee (approved):

Cargo.toml
MiguelDD1: +1 to use tags instead of `rev`.

PR #139, opened 2022-09-27
fix cors on http

kroos47:


Cargo.lock
```diff
@@ -320,7 +320,7 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "avail-light"
-version = "1.1.0"
```
aterentic-ethernal, line 323: Lets fix this in separate commit?

kroos47: yeh np!

aterentic-ethernal (approved):

src/http.rs
```diff
@@ -256,15 +256,21 @@ pub async fn run_server(
 		let counter_lock = counter_status.lock().unwrap();
 		status(&cfg, *counter_lock, db.clone())
 	});
+	let cors = warp::cors()
+		.allow_any_origin()
+		.allow_header("content-type")
+		.allow_methods(vec!["GET", "POST", "DELETE"]);
```
prabal-banerjee, line 262: We do not allow any POST or DELETE at the moment, but IMO we should not allow any origin for them. This legacy code, if remain unchanged, can be a security vulnerability later.

aterentic-ethernal: Never understood CORS implications completely :) This is only for user agents to allow requests to LC domain after OPTIONS call. I agree is better to have narrowest settings possible (like CORS setting per endpoint), but I don't understand what are vulnerabilities :)

PR #142, opened 2022-09-30
[DVT-113] add arm64 builds

rebelArtists:
**jira ticket:** 
https://polygon.atlassian.net/browse/DVT-113

**purpose:**
- add arm64 binary support for remote hosts (devnets)
- compress all output binaries

**tests:**
smoke tests, successful build action workflow run: https://github.com/maticnetwork/avail-light/actions/runs/3160544297

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,3 +52,6 @@ prometheus = { version = "0.13.0", default-features = false }
 proptest = "1.0.0"
 test-case = "1.2.3"
 
+[features]
+vendored = ["openssl-sys/vendored"]
+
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -47,6 +47,7 @@ warp = "0.3.2"
 tracing = "0.1.35"
 tracing-subscriber = { version = "0.3.15", features = ["json"] }
 prometheus = { version = "0.13.0", default-features = false }
+openssl = { version = "0.10", features = ["vendored"] }
```
rebelArtists, line 50: required, else run into following cross compile error for openssl: `run pkg_config fail: "pkg-config has not been configured to support cross-compilation.\n\nInstall a sysroot for the target platform and configure it via\nPKG_CONFIG_SYSROOT_DIR and PKG_CONFIG_PATH, or install a\ncross-compiling wrapper for pkg-config and set it via\nPKG_CONFIG environment variable."`

praetoriansentry: This seems to make the most sense I guess 
https://github.com/messense/rust-musl-cross/issues/35

praetoriansentry (approved):
lgtm

PR #143, opened 2022-10-05
Add commitments equality on app client.

aterentic-ethernal:


Cargo.toml
```diff
@@ -7,8 +7,8 @@ authors = ["Polygon Avail Team"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-kate-recovery = { git = "https://github.com/maticnetwork/avail.git", tag = "v1.1.0" }
-kate-proof = { git = "https://github.com/maticnetwork/avail.git", tag = "v1.1.0" }
+kate-recovery = { git = "https://github.com/maticnetwork/avail.git", rev = "b2427bf" }
```
MiguelDD1, line 10: Instead of `rev` I think we should use `tag`. 
Pls create a new tag (v1.1.1?) for `rev = b2427bf`.

MiguelDD1 (approved):

PR #144, opened 2022-10-10
add version check

kroos47:


src/main.rs
```diff
@@ -182,6 +182,15 @@ async fn do_main() -> Result<()> {
 
 	let rpc_url = rpc::check_http(&cfg.full_node_rpc).await?.clone();
 
+	let version = rpc::get_chain(&rpc_url).await?;
+	let runtime_version = rpc::get_runtime_version(&rpc_url).await?;
+	if version != "1.1.0"
+		&& runtime_version.spec_version != 5
+		&& runtime_version.spec_name != "data-avail"
+	{
+		panic!("version does not match the latest")
```
prabal-banerjee, line 191: Can we avoid panic and handle this error more gracefully?

src/main.rs
```diff
@@ -182,6 +182,15 @@ async fn do_main() -> Result<()> {
 
 	let rpc_url = rpc::check_http(&cfg.full_node_rpc).await?.clone();
 
+	let version = rpc::get_chain(&rpc_url).await?;
+	let runtime_version = rpc::get_runtime_version(&rpc_url).await?;
+	if version != "1.1.0"
```
prabal-banerjee, line 187: I may be wrong here, but if spec name is not `data-avail` but versions match, then this will not panic, right?

kroos47: yeh it will panic! or in if corrected it will exit

src/rpc.rs
```diff
@@ -307,3 +337,16 @@ pub fn cell_count_for_confidence(confidence: f64) -> u32 {
...
+
+pub async fn check_version(url: &str) -> Result<bool> {
+	let version = get_chain(url).await?;
+	let runtime_version = get_runtime_version(url).await?;
+	if version != "1.1.0"
+		|| runtime_version.spec_version != 5
+		|| runtime_version.spec_name != "data-avail"
+	{
+		return Err(anyhow!("Invalid version"));
+	} else {
```
aterentic-ethernal, line 349: Instead of having both network errors and invalid version under the Err, we probably should have `Ok(true)` for valid, `Ok(false)` for invalid, and `Err` for other cases, since client behavior would be different in all three cases?

aterentic-ethernal: Also, maybe it would be cool to have 
```
fn version_matches(version: String, runtime_version: RuntimeVersion) -> bool {
    version != "1.1.0" || runtime_version.spec_version != 5 || runtime_version.spec_name != "data-avail"
}```

And leave errors and side effects to upper layer:

```
let version = get_chain(url).await?;
let runtime_version = get_runtime_version(url).await?;
if version_matches(version, runtime_version) {
    ...do the thing
}```

Or even with proper typing:
```
let version = get_chain(url).await?;
let runtime_version = get_runtime_version(url).await?;
if version.not_below("1.1.0") && runtime_version.not_below(5, "data_avail") {
  // do the thing
}
```

src/main.rs
```diff
@@ -182,6 +182,14 @@ async fn do_main() -> Result<()> {
 
 	let rpc_url = rpc::check_http(&cfg.full_node_rpc).await?.clone();
 
+	match rpc::check_version(&rpc_url).await {
+		Ok(v) => v,
```
aterentic-ethernal, line 186: Do we use v here?

aterentic-ethernal (approved):

PR #152, opened 2022-11-17
identify protocol added

sh3ll3x3c:
- added identify protocol in order to solve the issue of bootstrap client storing the wrong port in the incoming client address
- added more swarm observability through trace and debug logs

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -565,7 +565,7 @@ pub fn init(
 		.map(|text| PreSharedKey::from_str(&text))
 		.transpose()?;
 	// create transport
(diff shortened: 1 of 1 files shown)


src/network.rs
```diff
@@ -446,7 +565,7 @@ pub fn init(
 		.map(|text| PreSharedKey::from_str(&text))
 		.transpose()?;
 	// create transport
-	let transport = setup_transport(id_keys, psk, port_reuse);
+	let transport = setup_transport(id_keys.clone(), psk, port_reuse);
```
aterentic-ethernal, line 568: setup_transport uses reference to key_pair internally so instead of cloning we can pass reference

PR #163, opened 2022-11-25
AV-80: Light clients dropping out of network when boostrap node goes offline

momosh-ethernal:
Added new network behaviors:

- Ping: as a network health check layer
- Keep alive: to prevent unwanted connection timeouts

Clearing the routing table from peers that closed connections, or from peers on the local network whose MDNS records expired.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -439,7 +439,21 @@ impl EventLoop {
 							.add_address(&peer_id, multiaddr);
 					}
 				},
-				_ => (),
+				MdnsEvent::Expired(addrs_list) => {
+					for (peer_id, multiaddr) in addrs_list {
+						debug!(
+							"MDNS got expired peer with ID: {:#?} and Address: {:#?}",
+							peer_id, multiaddr
+						);
+
+						if !self.swarm.behaviour_mut().mdns.has_node(&peer_id) {
+							self.swarm
+								.behaviour_mut()
+								.kademlia
+								.remove_address(&peer_id, &multiaddr);
+						}
+					}
+				},
 			},
 			swarm_event => {
 				// record Swarm events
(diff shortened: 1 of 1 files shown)


src/network.rs
```diff
@@ -413,7 +412,21 @@ impl EventLoop {
 							.add_address(&peer_id, multiaddr);
 					}
 				},
-				_ => (),
+				MdnsEvent::Expired(addrs_list) => {
+					for (peer_id, multiaddr) in addrs_list {
+						debug!(
```
aterentic-ethernal, line 417: Its also possible to write: `debug!("MDNS got expired peer with ID: {peer_id:#?} and Address: {multiaddr:#?}");`

aterentic-ethernal (approved):

src/network.rs
```diff
@@ -413,7 +412,21 @@ impl EventLoop {
...
+							"MDNS got expired peer with ID: {:#?} and Address: {:#?}",
+							peer_id, multiaddr
+						);
+
+						if !self.swarm.behaviour_mut().mdns.has_node(&peer_id) {
```
sh3ll3x3c, line 422: Should we also check whether we have active open connections?

src/network.rs
```diff
@@ -433,7 +446,19 @@ impl EventLoop {
...
 						cause,
 					} => {
-						trace!("Connection closed. PeerID: {:?}. Endpoint: {:?}. Num establ: {:?}. Cause: {:?}", peer_id, endpoint, num_established, cause);
+						trace!("Connection closed. PeerID: {:?}. Address: {:?}. Num establ: {:?}. Cause: {:?}", peer_id, endpoint.get_remote_address(), num_established, cause);
+
+						if let Some(cause) = cause {
+							match cause {
+								// remove peer with failed connection
+								ConnectionError::IO(_) | ConnectionError::Handler(_) => {
+									self.swarm.behaviour_mut().kademlia.remove_peer(&peer_id);
```
sh3ll3x3c, line 455: We should prob check for other open connections first, as there could be a case of partial failure when some connections are dropped and some are not...

src/network.rs
momosh-ethernal: because of the keep_alive behavior, the MDNS record will only expire on closed connections
this line also checks if the record is still active before remove

src/network.rs
momosh-ethernal: agree, but for the time being, until we implement a retry mechanism, this will have to suffice

sh3ll3x3c (approved):

PR #164, opened 2022-11-28
AV-79: Bootstrap node failing after a certain amount of time

momosh-ethernal:
There was a bug with the network_event output channel. It was overflowing because there was no receiver waiting to acquire the data after the stream was mapped once.
And that is exactly what happened in the case of the first node in the network, waiting to be dialed for its bootstrap to be able to complete.

We are now trying to send our events to all receivers and immediately removing them if send has failed.


src/network/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+		EventLoop::new(swarm, command_receiver, metrics, network_stream),
+	))
+}
+
+/// Read the pre-shared key file from the given directory
+fn get_psk(location: &String) -> Result<Option<String>> {
+	let path = Path::new(location);
+	match fs::read_to_string(path) {
+		Ok(text) => Ok(Some(text)),
+		Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(None),
```
aterentic-ethernal, line 106:
```suggestion
		Err([e](std::io::ErrorKind::NotFound)) => Ok(None),
```

src/network/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+use tracing::info;
+
+pub fn init(
+	seed: Option<u8>,
+	psk_path: &String,
+	metrics: Metrics,
+	port_reuse: bool,
+) -> Result<(Client, Arc<NetworkStream>, EventLoop)> {
+	// Create a public/private key pair, either based on a seed or random
+	let id_keys = match seed {
```
aterentic-ethernal, line 45: Ideal for if let..else from rust 1.65 :D

src/network/stream.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+pub struct NetworkStream {
+	stream: Mutex<EventStream>,
+}
+
+impl NetworkStream {
+	pub fn new() -> Self {
+		Self {
+			stream: Mutex::new(EventStream {
```
aterentic-ethernal, line 35: I think (hope) we can derive default on EventStream maybe even on NetworkStream and have only Self::default() or something like that.

sh3ll3x3c (approved):

src/network/mod.rs
momosh-ethernal: not possible, due to:
`mismatched types: expected struct `std::io::Error`, found enum `std::io::ErrorKind`

aterentic-ethernal (approved):

PR #165, opened 2022-11-29
AV-83: Add a signaling mechanism for successfully completing the bootstrap process

momosh-ethernal:


src/main.rs
```diff
@@ -205,13 +204,13 @@ async fn do_main() -> Result<()> {
...
 			.context("Connection is not established")?;
-		bootstrap_nodes = vec![node];
-		network_client.bootstrap(bootstrap_nodes).await?;
+		vec![node]
 	} else {
-		// Now that we have something to bootstrap with, just do it
-		info!("Bootstraping the DHT with bootstrap nodes...");
-		network_client.bootstrap(bootstrap_nodes).await?;
-	}
+		bootstrap_nodes
```
sh3ll3x3c, line 209: We should add a status debug log here, same as it was before.

sh3ll3x3c (approved):

PR #167, opened 2022-11-29
add threshold to DHT fetching 

kroos47:


src/app_client.rs
```diff
@@ -72,6 +72,7 @@ async fn process_block(
 	app_id: u32,
 	block: &ClientMsg,
 	pp: PublicParameters,
+	thresh: usize,
```
aterentic-ethernal, line 75: We can put threshold into AppClientConfig struct, to avoid cluttering parameter list.

src/types.rs
```diff
@@ -412,6 +412,8 @@ pub struct RuntimeConfig {
 	pub max_cells_per_rpc: Option<usize>,
 	/// Time-to-live for DHT entries in seconds (default: 3600).
 	pub ttl: Option<u64>,
+	/// Threshold for the number of cells fetched from DHT,
+	pub threshold: usize,
```
aterentic-ethernal, line 416: I think it would be good to have default value in case no option is specified in config, to avoid crashing LCs with old config. Like `dht_parallelization_limit`, for example.

aterentic-ethernal (approved):

PR #171, opened 2022-12-06
Expose Kademlia parameters through config

sh3ll3x3c:
- 10 new Kademlia parameters have been exposed through LC config file
- Supplemented README file
- Modified several default Kademlia parameters to make the more in line with rust-libp2p defaults

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -157,7 +157,7 @@ async fn do_main() -> Result<()> {
 		&cfg.libp2p_psk_path,
 		libp2p_metrics,
 		cfg.libp2p_tcp_port_reuse,
-		cfg.replication_factor
+		(&cfg).into(),
 	)
 	.context("Failed to init Network Service")?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -157,6 +157,7 @@ async fn do_main() -> Result<()> {
 		&cfg.libp2p_psk_path,
 		libp2p_metrics,
 		cfg.libp2p_tcp_port_reuse,
+		cfg.replication_factor
 	)
 	.context("Failed to init Network Service")?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,7 @@ use crate::{
 mod app_client;
 mod consts;
 mod data;
+mod defaults;
 mod http;
 mod light_client;
 mod network;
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -157,6 +157,7 @@ async fn do_main() -> Result<()> {
 		&cfg.libp2p_psk_path,
 		libp2p_metrics,
 		cfg.libp2p_tcp_port_reuse,
+		(&cfg).into(),
```
aterentic-ethernal, line 160: Since we are passing struct to `network::init`, can we remove other parameters and use them from struct?

src/network/mod.rs
```diff
@@ -59,12 +62,35 @@ pub fn init(
 	// create swarm that manages peers and events
 	let swarm = {
 		let mut kad_cfg = KademliaConfig::default();
-		kad_cfg.set_query_timeout(Duration::from_secs(5 * 60));
+		kad_cfg
+			.set_publication_interval(Some(Duration::from_secs(
+				kad_config.publication_interval as u64,
+			)))
+			.set_replication_interval(Some(Duration::from_secs(
```
aterentic-ethernal, line 69: Maybe we can make conversion during serde, so we can have valid duration here and avoid mixing conversions? Or at least to move it at `From` implementation for `KademliaConfig`?

src/network/mod.rs
```diff
@@ -59,12 +62,35 @@ pub fn init(
...
+				kad_config.record_replication_interval as u64,
```
aterentic-ethernal, line 70: I think we can use `kad_config.record_replication_interval.into()` (and on other places of conversion) to preserve compile time safety.

src/types.rs
```diff
@@ -336,6 +336,10 @@ mod port_range_format {
 	}
 }
 
+fn default_libp2p_psk_path() -> String {
```
aterentic-ethernal, line 339: wdyt of moving defaults to defaults module and then use it as `defaults::libp2p_psk_path`?

src/main.rs
sh3ll3x3c: Well not for now, but in general - absolutely. The Kad parameters are logically separate from libp2p parameter, such as psk path and port reuse. In some next refactor iteration we'll def create an overarching network config structure that'll contain all of those...

luka-ethernal (approved):

PR #179, opened 2022-12-21
AV-120: Insert rows into dht

aterentic-ethernal:
This PR adds rows insertion into DHT, with additional metrics for tracking rows insertion success rate and timings. Rows are inserted into DHT if client is configured with partition, after partition is fetched from DHT. Dependency to `avail-core` will be updated after tagging.

### DIFF
--- a/src/light_client.rs
+++ b/src/light_client.rs
@@ -35,7 +35,7 @@ use codec::Encode;
 use dusk_plonk::commitment_scheme::kzg10::PublicParameters;
 use futures::future::join_all;
 use kate_recovery::{
-	commitments,
+	commitments, data,
 	matrix::{Dimensions, Position},
 };
 use rocksdb::DB;
@@ -129,7 +129,7 @@ pub async fn run(
 			    continue;
 			};
 
-			if !(dimensions.cols() > 2) {
+			if dimensions.cols() <= 2 {
 				error!(block_number, "more than 2 columns is required");
 			}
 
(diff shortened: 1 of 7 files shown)


src/light_client.rs
```diff
@@ -268,8 +268,23 @@ pub async fn run(
 							.collect::<Vec<_>>();
 						rpc_fetched.extend(partition_fetched_filtered.clone());
 					}
+
+					// TODO: Skip erasure coded rows from rpc_fetched
+					// e.g. let rpc_fetched_data = rpc_fetched.iter().filter(...)
+					let rpc_fetched_data_rows = data::rows(&dimensions, &rpc_fetched);
```
sh3ll3x3c, line 274: Should the LC be responsible for row insertions? Or does it make more sense for the app client to do it, because the rows are being used solely by the app client?

src/network/client.rs
```diff
@@ -42,6 +45,23 @@ impl DHTCell {
 	}
 }
 
+struct DHTRow((RowIndex, Vec<u8>));
+
+impl DHTRow {
+	fn reference(&self, block: u32) -> String {
+		self.0 .0.reference(block)
```
sh3ll3x3c, line 52: Is this a unique ref per block?

sh3ll3x3c (approved):

src/light_client.rs
aterentic-ethernal: Its actually "fat" client, it will push rows from partition if partition is configured. Maybe it could be additional configuration, but maybe it makes sense to push everything upfront.

src/network/client.rs
aterentic-ethernal: Row reference is block_number:row_number, it should be unique. Once we have proper type, I guess monstrous constructions like `self.0.0` will disappear.

momosh-ethernal (approved):

PR #186, opened 2023-01-01
fix: make sure config defaults are properly implemented

prabal-banerjee:
The README suggests default values for the configurable parameters. Due to code changes, some of them are not enforced or wrongly enforced in `v1.3.0`. This PR aims to fix that. 

In the latest main, new parameters have been introduced and their defaults are captured in a separate file. This PR aims to follow a uniform convention by bringing all of them within the `RuntimeConfig` Default and enforce them through the updated `#[serde(default)]` directive.


src/types.rs
```diff
@@ -454,19 +430,19 @@ impl Default for RuntimeConfig {
...
+			query_timeout: 60,
+			query_parallelism: 3,
+			caching_max_peers: 1,
 			disjoint_query_paths: false,
-			max_kad_record_number: defaults::max_kad_record_number(),
-			max_kad_record_size: defaults::max_kad_record_size(),
-			max_kad_provided_keys: defaults::max_kad_provided_keys(),
+			max_kad_record_number: 2400000,
+			max_kad_record_size: 100,
+			max_kad_provided_keys: 1024,
```
aterentic-ethernal, line 445: Marking specific field as default and providing default function enables omitting that field in config file (and reusing it in default struct value). Does this code preserve that behavior? And if You know, how is doing that? I guess default struct value needs to be instantiated to use some value from it (in case when one field is missing for example).

aterentic-ethernal: @luka-ethernal showed me docs that explains that behavior. Additional considerations could be how to enforce mandatory values, and what to do in case of optionals (e.g. if field is omitted, user could expect None instead of default value which could be different).

sh3ll3x3c:
In order to preserve the mandatory field behavior, we should probably remove mandatory fields from the Default implementation.

src/types.rs
prabal-banerjee: > Additional considerations could be how to enforce mandatory values, and what to do in case of optionals (e.g. if field is omitted, user could expect None instead of default value which could be different).
Agreed. 
For optionals, does `confy` support them? Could not find anything in the doc which suggests so. It requires every field to have Default implemented. 
For the defaults, I have tried to capture all of them in the README so that the second confusion could be avoided as much as possible.

prabal-banerjee:
@sh3ll3x3c My naive attempt did not work. It requires all the fields to have default implemented. But there might be some directive I'm missing.

src/types.rs
aterentic-ethernal: Lol I was able to post second comment but wasnt able to see yours reply before refreshing page, thanks github. Makes sense, we will just briefly try to cover mandatory and optional cases.

aterentic-ethernal: After discussion it seems that mandatory field case can be avoided which will reduce complexity and that optional cases can be handled separately so Ill resolve my comment.

luka-ethernal:
Sorry, accidentally removed alex's approval :(

luka-ethernal (approved):

PR #189, opened 2023-01-26
Return decoded data on appdata endpoint.

aterentic-ethernal:
This PR adds `decode` boolean query parameter to `appdata` HTTP endpoint, to support retrieving decoded submitted data. 

If `decode` query parameter is missing or `false`, endpoint behaves same as before, preserving backward compatibility, and return hex string of encoded extrinsic. If `decode` is set to true, extrinsic is decoded and base64 string of submitted data is returned.

There was several options for specifying this behavior:
- config parameter: this is less flexible and it doesn't preserve compatibility if client is started with different setting
- different endpoint: since schema is same, but meaning of extrinsic field is same, same endpoint seems more natural
- specifying accept header: this allows for content negotiation but devs are less familiar with this kind of api
- query parameter: similar to accept header but more familiar
 
Base64 is chosen as most common encoding and easier use with online tools and in browser, but there could be more compact options which would require specific endpoint for each extrinsic, or some other solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,17 +183,30 @@ Result:
 > `blockNumber << 32 | int32(confidence * 10 ** 7)`, where confidence is represented out of 10 ** 9.
 
 
-Given a block number (as _(hexa-)_ decimal number), return the content of the data if specified in config in hex string format
+Given a block number (as _(hexa-)_ decimal number), return the extrinsic in hex string format, if app id is specified in the config
 
 ```bash
-curl -s localhost:7000/v1/appdata/ <block-number>
+curl -s localhost:7000/v1/appdata/<block-number>
 ```
 
 Result:
 
 ```json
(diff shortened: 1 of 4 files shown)


luka-ethernal (approved):

README.md
```diff
@@ -192,9 +192,22 @@ curl -s localhost:7000/v1/appdata/ <block-number>
 Result:
 
 ```json
-{"block":386,"extrinsics":[{"app_id":1,"signature":{"Sr25519":"be86221cc07a461537570637d75a0569c2210286e85c693e3b31d94211b1ef1eaf451b13072066f745f70801ad6af0dcdf2e42b7bf77be2dc6709196b4d45889"},"data":"0x313537626233643536653339356537393237633664"}]}
+{"block":386,"extrinsics":["{hex_encoded_data}"]}
```
prabal-banerjee, line 195: IMO, the prev example was better as it showed that it included signature as well as the hex encoded data. Also to note, IIRC, the hex encoded data was not the submitted data but the extrinsic calldata.

prabal-banerjee (approved):

README.md
aterentic-ethernal: You are right, its a bad wording. But what to put as example since it returns hex encoded extrinsic and you have to decode it to get scheme as in this example?

PR #197, opened 2023-02-14
adding appdata to sync_client

kroos47:


aterentic-ethernal (approved):
I think we should pay attention to load increase between sync clients without and with app part (it should be acceptable). And also, at some point lets rename ClientMsg to BlockVerifiedMessage or BlockVerifiedEvent or just BlockVerified :)

PR #199, opened 2023-02-28
AV-135: Implement relaying logic with flag for an LC

momosh-ethernal:
Implemented NAT traversal, through the use of relays and DCUtR protocol

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -173,14 +173,29 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 	} else {
 		cfg.libp2p_port.0
 	};
+
+	// always listen on UDP to prioritize QUIC
 	network_client
 		.start_listening(
 			Multiaddr::empty()
-				.with("0.0.0.0".parse::<Ipv4Addr>().unwrap().into())
-				.with(Protocol::Tcp(port)),
+				.with(Protocol::from(Ipv4Addr::UNSPECIFIED))
+				.with(Protocol::Udp(port))
+				.with(Protocol::QuicV1),
 		)
 		.await
-		.context("Listening not to fail.")?;
+		.context("Listening on UDP not to fail.")?;
+	// if there were no provided relays, that means we're one of those
+	// than we need to act as a Relay, listen on TCP also
+	if cfg.relays.is_empty() {
+		network_client
+			.start_listening(
+				Multiaddr::empty()
+					.with(Protocol::from(Ipv4Addr::UNSPECIFIED))
+					.with(Protocol::Tcp(port)),
+			)
+			.await
+			.context("Listening on TCP not to fail.")?;
+	}
 
 	// Check if bootstrap nodes were provided
 	let bootstrap_nodes = cfg
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,7 +11,7 @@ use anyhow::{anyhow, Context, Result};
 use async_std::stream::StreamExt;
 use avail_subxt::primitives::Header;
 use consts::STATE_CF;
-use libp2p::{metrics::Metrics as LibP2PMetrics, Multiaddr, PeerId};
+use libp2p::{metrics::Metrics as LibP2PMetrics, multiaddr::Protocol, Multiaddr, PeerId};
 use prometheus_client::registry::Registry;
 use rand::{thread_rng, Rng};
 use rocksdb::{ColumnFamilyDescriptor, Options, DB};
@@ -174,7 +174,11 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 		cfg.libp2p_port.0
 	};
 	network_client
-		.start_listening(format!("/ip4/0.0.0.0/tcp/{}", port).parse()?)
+		.start_listening(
+			Multiaddr::empty()
+				.with("0.0.0.0".parse::<Ipv4Addr>().unwrap().into())
+				.with(Protocol::Tcp(port)),
+		)
 		.await
 		.context("Listening not to fail.")?;
 
--- a/src/network/event_loop.rs
+++ b/src/network/event_loop.rs
@@ -18,8 +18,7 @@ use libp2p::{
 	},
 	identify::{Event as IdentifyEvent, Info},
 	kad::{
-		BootstrapOk, GetRecordOk, InboundRequest, KademliaEvent, PeerRecord, PutRecordOk, QueryId,
-		QueryResult,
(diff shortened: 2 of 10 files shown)


src/network/event_loop.rs
```diff
@@ -165,12 +114,54 @@ impl EventLoop {
...
+				itertools::Either<
+					itertools::Either<
+						itertools::Either<
+							itertools::Either<
+								itertools::Either<
+									itertools::Either<
+										itertools::Either<
+											ConnectionHandlerUpgrErr<
+												itertools::Either<
+													InboundHopFatalUpgradeError,
```
luka-ethernal, line 127: wow :)

luka-ethernal (changes requested):

src/network/event_loop.rs
aterentic-ethernal: Maybe we can replace qualified path with use for `Either` and hide complexity a little :S

luka-ethernal (approved):

src/network/event_loop.rs
momosh-ethernal: it's getting more and more beautiful

src/network/event_loop.rs
```diff
@@ -365,6 +370,55 @@ impl EventLoop {
 						"AutoNat Old status: {:#?}. AutoNat New status: {:#?}",
 						old, new
 					);
+					// check if went private or are private
+					// if so, create reservation request with relay
+					if new == NatStatus::Private {
+						info!("Autonat says we're still private.");
+						// choose relay by random
+						if let Some(relay) = self.relay_nodes.choose(&mut rand::thread_rng()) {
```
sh3ll3x3c, line 378: Should we always select a different relay or should we rely the same ones? Keeping in mind that autonat can return and will return NAT status private very often in real use cases...

src/network/event_loop.rs
```diff
@@ -365,6 +370,55 @@ impl EventLoop {
...
+							let (relay_id, relay_addr) = relay.clone();
+							// appoint this relay as our chosen one
+							self.relay_reservation(relay_id, relay_addr);
+						}
+					};
+				},
+			},
+			SwarmEvent::Behaviour(BehaviourEvent::Relay(event)) => match event {
```
sh3ll3x3c, line 386: Are these all relay events (to be extracted to a separate relay repo at some point in the near future) or client ones?

src/network/event_loop.rs
```diff
@@ -510,4 +565,20 @@ impl EventLoop {
 			},
 		}
 	}
+
+	fn reset_relay_reservation(&mut self) {
+		self.relay_reservation = RelayReservation {
+			id: PeerId::random(),
```
sh3ll3x3c, line 571: Should the ID here be random or empty?

src/network/event_loop.rs
```diff
@@ -299,28 +271,60 @@ impl EventLoop {
...
+						if peer_id == self.relay_reservation.id
+							&& self.relay_reservation.is_reserved == false
 						{
-							for addr in listen_addrs {
-								self.swarm
-									.behaviour_mut()
-									.kademlia
-									.add_address(&peer_id, addr);
+							match self.swarm.listen_on(
+								self.relay_reservation
```
sh3ll3x3c, line 284: What does the relay reservation actually do here?

src/network/mod.rs
```diff
@@ -75,50 +87,128 @@ pub fn init(
...
+	let noise_keys = Keypair::<X25519Spec>::new().into_authentic(&id_keys)?;
+
+	// Create Transport
+	// relay transport configuration used in relay clients
+	let local_peer_id = PeerId::from(id_keys.public());
+	let (relay_client_transport, relay_client_behaviour) = relay::client::new(local_peer_id);
+
+	let mut yamux_config = YamuxConfig::default();
+	// enabling propper flow-control:
+	// window updates are only sent when buffered data has been consumed
```
sh3ll3x3c, line 99: If the receiver is too slow, in our case LC is bogged down for whatever reason, this could potentially cause back-pressure to slow down the client that's sending the data e.g. the fat client. 

It might be better to set it to `on_receive()`. wdyt?

src/network/mod.rs
```diff
@@ -75,50 +87,128 @@ pub fn init(
...
+	yamux_config.set_window_update_mode(WindowUpdateMode::on_read());
+
+	let quic_transport = {
```
sh3ll3x3c, line 102: Are we using both QUIC and TCP transports by default in clients (not relays)?

src/main.rs
```diff
@@ -174,7 +174,11 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 		cfg.libp2p_port.0
 	};
 	network_client
-		.start_listening(format!("/ip4/0.0.0.0/tcp/{}", port).parse()?)
+		.start_listening(
+			Multiaddr::empty()
+				.with("0.0.0.0".parse::<Ipv4Addr>().unwrap().into())
```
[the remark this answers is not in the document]
momosh-ethernal: true dat, this should be fixed

src/network/event_loop.rs
momosh-ethernal: since we don't have proper discovery mechanisms in place, a random relay is preferred to kinda relax traffic going through relays

once we have discovers, we should connect the "closest" ones

src/network/event_loop.rs
momosh-ethernal: separations will happen
anything that isn't a relay should have its client events, and relays should only be concerned with their own events

src/network/event_loop.rs
momosh-ethernal: this is just for resetting it to something different other than current ID

I need to check, not sure but I think that PeerId doesn't implement that
will check and fix it, if possible

src/network/event_loop.rs
momosh-ethernal: stops node from grabbing more reservations

src/network/mod.rs
momosh-ethernal: that mode introduces a possibility for deadlock
we should discuss this further

src/network/mod.rs
momosh-ethernal: yes, both are needed
quic is prioritized, but in case of NATs we need TCP for relaying

src/network/event_loop.rs
aterentic-ethernal: also it seems that handle_event is not async and we can remove it

PR #209, opened 2023-05-12
Luka/check finality

luka-ethernal:


src/subscriptions.rs
```diff
@@ -1,41 +1,316 @@
...
+		async move {
+			while let Some(Ok(header)) = header_subscription.next().await {
+				let received_at = Instant::now();
+				let head_hash: H256 = Encode::using_encoded(&header, blake2_256).into();
+				msg_sender
+					.send(Messages::NewHeader(header, received_at))
+					.expect("Receiver should not be dropped.");
+
+				// Fetch all events at the incoming header hight.
+				let events = subxt_client
```
prabal-banerjee, line 147: The total events per block can be quite big. Is there a more optimized way of fetching pallet based events?

src/subscriptions.rs
```diff
@@ -1,41 +1,316 @@
...
+					.expect("Fetching events should succeed.");
+
+				// Filter out just new authorities event.
+				let new_auths =
+					events.find_last::<avail_subxt::api::grandpa::events::NewAuthorities>();
+
+				// If the event exists, send the new auths over the message channel.
+				if let Ok(Some(auths)) = new_auths {
+					// Fetch set ID at the incoming header hight (needed to verify justification).
+					let set_id = subxt_client
```
prabal-banerjee, line 160: Can the full node falsify this information? Is the storage element checked against a root in the header (or something similar?)

prabal-banerjee: Same question for many other RPC calls below.

prabal-banerjee (approved):

src/subscriptions.rs
luka-ethernal: I haven't found a way, so far. All the other API's either supply authority keys in a different format, requiring more RPC's or aren't working the way they should.

luka-ethernal:
So, @prabal-banerjee, I changed the way validator set changes are acquired, which is now directly from header, so it can be checked agains the very hash of the block. That block necessarily has justifications with previous validator set, so it is automatically checked. As a bonus, we no longer fetch the entire events list for a block, saving us a pretty big RPC call as well.
The only thing remaining is the set ID, which still is retrieved from storage, I might also look into investigating removing that.

prabal-banerjee:
This is awesome!

About ID, how is ID set? Is it monotonically increasing in every era? I have a hunch that the entire finality check should be doable without querying state (or relying on full node), in some way. This is bcoz Polkadot always had publicly stated how they want LCs to be first class citizens. Fetching the ID from state is not a deal-breaker, ofcourse, but being able to avoid it would be a nice to have. Hope you got my point.

luka-ethernal:
Yes, I have come to the same conclusion, it looks like it's just monotonically incremented on every change, but I don't know if there are edge cases. In any way, I'm gonna test it on a devnet chain and if it is always just incremented, I think it would be enough (together with the reference you provided).

PR #217, opened 2023-05-29
Implement Kademlia memory store

sh3ll3x3c:
- default Kademlia memory store replaced with our own implementation
- [Memory optimization] - fat clients are now removing all local records on PUT operations
- [Memory optimization] - fat clients are no longer storing incoming records

### DIFF
--- a/src/network/event_loop.rs
+++ b/src/network/event_loop.rs
@@ -194,12 +194,18 @@ impl EventLoop {
 						QueryResult::PutRecord(result) => {
 							if let Some(v) = self.pending_kad_query_batch.get_mut(&id) {
 								*v = Some(result.map_err(Into::into).map(|ok_res| {
+									// Remove local records for fat clients (memory optimization)
 									if self.kad_remove_local_record {
 										self.swarm
 											.behaviour_mut()
 											.kademlia
 											.remove_record(&ok_res.key);
 									};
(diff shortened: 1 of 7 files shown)


src/network/event_loop.rs
```diff
@@ -194,12 +194,18 @@ impl EventLoop {
...
 							if let Some(v) = self.pending_kad_query_batch.get_mut(&id) {
 								*v = Some(result.map_err(Into::into).map(|ok_res| {
+									// Remove local records for fat clients (memory optimization)
 									if self.kad_remove_local_record {
 										self.swarm
 											.behaviour_mut()
 											.kademlia
 											.remove_record(&ok_res.key);
 									};
+									self.swarm
```
momosh-ethernal, line 204: always shrinking on each put? without any checks?

sh3ll3x3c: not ideal, a better solution would be to shrink it on every block, but the call itself is very resource cheap even with this approach - I'm seeing no performance penalties under a load test...

aterentic-ethernal: Maybe its easy to invoke it on batch complete?

momosh-ethernal (approved):

aterentic-ethernal (approved):

PR #219, opened 2023-05-30
Cleanup after Relay and Bootstrap node separation 

momosh-ethernal:


src/main.rs
```diff
@@ -243,9 +231,9 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 	let last_full_node_ws = data::get_last_full_node_ws_from_db(db.clone())?;
 
 	let version = rpc::Version {
-		version: "1.6.0".to_string(),
+		version: cfg.full_node_rpc_version.clone(),
```
aterentic-ethernal, line 234: This shouldn't be configuration since its compile time dependency. But, since we are using our avail client to communicate with the node, maybe specifying dependency is not needed any more (its implicitly specified with avail-subxt version)?

src/types.rs
```diff
@@ -197,24 +197,40 @@ pub struct RuntimeConfig {
 	/// If set to seed, keypair will be generated from that seed.
 	/// If set to key, a valid ed25519 private key must be provided, else the client will fail
 	/// If `secret_key` is not set, random seed will be used.
-	pub secret_key: Option<SecretKey>,
+	pub libp2p_secret_key: Option<SecretKey>,
```
sh3ll3x3c, line 200: We need to make sure the example config is also changes accordingly.

src/network/mod.rs
```diff
@@ -88,57 +85,21 @@ pub fn init(
...
-				Either::Right((peer_id, connection)) => (peer_id, StreamMuxerBox::new(connection)),
-			},
-		))?
-		.boxed()
-	} else {
-		// but, if we're a client that will use Relaying
-		// we need to mix relay client transport with TCP
-		let relay_client_tcp_transport = OrTransport::new(relay_client_transport, tcp_transport)
-			.upgrade(Version::V1)
+	let transport = {
```
sh3ll3x3c, line 91: Do we have a TCP fallback if QUIC is not possible?

src/main.rs
luka-ethernal: Agree! I think avail-sbuxt is enough to guarantee compatibility right now.

src/main.rs
```diff
@@ -50,12 +50,20 @@ use tikv_jemallocator::Jemalloc;
...
+	author = "Avail Team",
+	version = "0.1.0"
+)]
+struct CliOpts {
+	#[structopt(
+		short = "c",
+		long = "config",
+		default_value = "config.yaml",
+		help = "yaml configuration file"
+	)]
```
aterentic-ethernal, line 66: This should be reverted, right version is replaced with old one.

PR #221, opened 2023-05-31
Introduce additional p2p network observability

sh3ll3x3c:
- Added a new command for dumping network stats on each new block
- Added Kademlia routing table connected peer list dump
- Added new metric for a number of total connected peers from the routing table

### DIFF
--- a/src/telemetry/metrics.rs
+++ b/src/telemetry/metrics.rs
@@ -6,6 +6,7 @@ use prometheus_client::{
 };
 
 // Defines metrics struct that is used with Prometheus for light client.
+#[derive(Clone)]
 pub struct Metrics {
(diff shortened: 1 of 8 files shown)


src/telemetry/metrics.rs
```diff
@@ -32,6 +34,7 @@ pub enum MetricEvent {
 	DHTPutSuccess(f64),
 	DHTPutRowsDuration(f64),
 	DHTPutRowsSuccess(f64),
+	KadRoutinTablePeerNum(u32),
```
aterentic-ethernal, line 37: Typo?

sh3ll3x3c: Absolutely

PR #223, opened 2023-06-01
Luka/check finality sync

luka-ethernal:


Cargo.toml
```diff
@@ -8,7 +8,7 @@ authors = ["Avail Team"]
 
 [dependencies]
 kate-recovery = { version = "0.8", git="https://github.com/availproject/avail-core", tag = "da-primitives/v0.4.6" }
-avail-subxt = { git = "https://github.com/availproject/avail.git", tag = "v1.6.0" }
```
aterentic-ethernal, line 11: Does it make sense to depend only on tagged versions of avail, so we are confident in which node version should work? That means we have to enforce tagging when changes are on both LC and the node, but it seems as good trade off.

luka-ethernal: Yes, but there's one commit I'm dependent on, which is just after the tag and I don't know what's the proper way to handle this. Tag 1.6.1? Or something else?

aterentic-ethernal: Actually if that is unstable version on develop branch which didn't landed into the main, we cannot do something meaningful. But in that case LC is also in unstable state. Soooo :man_shrugging:

src/subscriptions.rs
```diff
@@ -1,40 +1,279 @@
...
+							.then(|| precommit.clone().id)
+							.ok_or(anyhow!("Not signed by this signature!"))
+					})
+					.collect::<Result<Vec<_>>>();
+				let Ok(signer_addresses) = signer_addresses else {
+					break 'mainloop Err(signer_addresses.unwrap_err());
+				};
+
+				// Match all the signer addresses to the current validator set.
+				let num_matched_addresses = signer_addresses
```
sh3ll3x3c, line 243: Are the validators the only ones able to sign justifications?

src/sync_client.rs
```diff
@@ -256,31 +263,93 @@ pub async fn run(
...
 ) {
+	let rpc_client = sync_client.get_client();
 	if sync_blocks_depth >= 250 {
 		warn!("In order to process {sync_blocks_depth} blocks behind latest block, connected nodes needs to be archive nodes!");
 	}
+
+	// Fetch validator set at current height
+	let final_validator_set = rpc::get_valset_by_block_number(&rpc_client, end_block)
+		.await
+		.unwrap();
```
sh3ll3x3c, line 274: Should we panic inside the sync client or handle these errors differently?

src/sync_client.rs
```diff
@@ -256,31 +263,93 @@ pub async fn run(
...
+
+	// Fetch the set ID from storage at current height
+	let final_set_id = rpc::get_set_id_by_block_number(&rpc_client, end_block)
+		.await
+		.unwrap();
```
sh3ll3x3c, line 279: ditto

src/utils.rs
```diff
@@ -40,6 +42,27 @@ pub(crate) fn extract_app_lookup(extension: &HeaderExtension) -> &DataLookup {
 	}
 }
 
+pub fn filter_auth_set_changes(header: DaHeader) -> Vec<Vec<(AuthorityId, u64)>> {
```
sh3ll3x3c, line 45: Do we need a test for this func?

src/sync_client.rs
```diff
@@ -256,31 +263,93 @@ pub async fn run(
...
-					error!(block_number, "Cannot process block: {error:#}");
-				}
-			},
-		)
-		.await;
+		// TODO: Should we handle unprocessed blocks differently?
+		if let Err(error) =
+			process_block(&sync_client, block_number, cfg_clone, &pp, block_tx).await
+		{
+			error!(block_number, "Cannot process block: {error:#}");
```
sh3ll3x3c, line 337: Maybe we should create a list of skipped blocks and save it on disk?

src/subscriptions.rs
```diff
@@ -1,40 +1,279 @@
...
+
+	// Forming a channel for sending any relevant events gathered asynchronously through Substrate WS API.
+	let (msg_sender, mut msg_receiver) = unbounded_channel::<Messages>();
+
+	// Task that produces headers and new validator sets
+	tokio::spawn({
+		// let subxt_client = subxt_client.clone();
+		let msg_sender = msg_sender.clone();
+		async move {
+			while let Some(Ok(header)) = header_subscription.next().await {
```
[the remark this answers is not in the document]
luka-ethernal: Well, we probably do, but it's a bit out of scope. We basically rely on this subscription for all of LC's lifetime. It's the entire driving mechanism for LC. This was actually done in the same manner before, and on any error, we simply die and rely on external mechanism for restart. Of course, we need to make it more resilient, but it's a much bigger task.

src/subscriptions.rs
```diff
@@ -1,40 +1,279 @@
...
+) -> Result<()> {
+	let mut header_subscription = subxt_client
+		.rpc()
+		.subscribe_finalized_block_headers()
+		.await?;
+	// Get the hash of the head (finalized)
+	let last_finalized_block_hash = rpc::get_chain_head_hash(&subxt_client).await?;
+
+	// Current set of authorities, implicitly trusted, fetched from grandpa runtime.
+	let mut validator_set =
```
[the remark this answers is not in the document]
luka-ethernal: Yes and no. We change nodes on startup in case there are multiple. But also, changes in the validator set are part of a header and that header is always signed off by previous validator set, so if we assume implicit trust in the first validator set (the one active on startup), we don't need to trust the node anymore.

src/sync_client.rs
luka-ethernal: Yes, I will modify sync client to act more like LC, with better error handling. I meant to do that anyway, but forgot.

src/utils.rs
luka-ethernal: Well, maybe in avail-subxt, or similar. Creating a fake header here with random digest for testing seems odd.

src/sync_client.rs
luka-ethernal: see above

src/subscriptions.rs
```diff
@@ -1,40 +1,279 @@
...
+	}
+}
+
+#[derive(Clone, Debug)]
+enum Messages {
+	Justification(GrandpaJustification),
+	ValidatorSetChange((Vec<EdPublic>, u64)),
+	NewHeader(Header, Instant),
+}
+async fn subscribe_check_and_process(
```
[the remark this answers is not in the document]
luka-ethernal: Will do.

src/sync_client.rs
luka-ethernal: At the second look, I think we should panic. Whereas we can have errors processing blocks themselves, without failing, errors in checking the validity should be fatal. I will of course add some more context, but I think it should stay.

sh3ll3x3c (approved):

PR #229, opened 2023-06-26
Periodic Bootstraps

momosh-ethernal:
This PR adds periodic bootstraps, as per libp2p KAD [specification](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#bootstrap-process) Intervals are part of the LC configuration.

Other changes:

- keep track of the bootstrap process through implemented state struct
- refactored relay reservation process
- refactored KAD Put Batch command to support multiple async commands

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
+.*
 /target
 **/*.rs.bk
-.env
-.DS_Store
 config.yaml
 avail_light_store
--- a/README.md
+++ b/README.md
@@ -81,29 +81,31 @@ http_server_port = "7000"
 # If set to seed, keypair will be generated from that seed.
 # If set to key, a valid ed25519 private key must be provided, else the client will fail
 # If `secret_key` is not set, random seed will be used.
-secret_key = { key =  "1498b5467a63dffa2dc9d9e069caf075d16fc33fdd4c3b01bfadae6433767d93" }
-# Libp2p service port range (port, range) (default: 37000).
-libp2p_port = "37000"
-# Configures LibP2P TCP port reuse for local sockets, which implies reuse of listening ports for outgoing connections to enhance NAT traversal capabilities (default: false)
-libp2p_tcp_port_reuse = false
-# Configures LibP2P AutoNAT behaviour to reject probes as a server for clients that are observed at a non-global ip address (default: false)
-libp2p_autonat_only_global_ips = false
-# Libp2p AutoNat throttle period for re-using a peer as server for a dial-request. (default: 1 sec)
-libp2p_autonat_throttle = 1
+secret_key = { seed="1" }
+# P2P service port (default: 37000).
+port = 3700
+# Configures TCP port reuse for local sockets, which implies reuse of listening ports for outgoing connections to enhance NAT traversal capabilities (default: false)
+tcp_port_reuse = bool
+# Configures AutoNAT behaviour to reject probes as a server for clients that are observed at a non-global ip address (default: false)
+autonat_only_global_ips = false
+# AutoNat throttle period for re-using a peer as server for a dial-request. (default: 1 sec)
+autonat_throttle = 2
 # Interval in which the NAT status should be re-tried if it is currently unknown or max confidence was not reached yet. (default: 10 sec)
-libp2p_autonat_retry_interval = 10
+autonat_retry_interval = 10
 # Interval in which the NAT should be tested again if max confidence was reached in a status. (default: 30 sec)
-libp2p_autonat_refresh_interval = 30
-# Libp2p AutoNat on init delay before starting the fist probe. (default: 5 sec)
-libp2p_autonat_boot_delay = 5
-# Sets libp2p application-specific version of the protocol family used by the peer. (default: "/avail_kad/id/1.0.0")
-libp2p_identify_protocol = "/avail_kad/id/1.0.0"
-# Sets libp2p agent version that is sent to peers. (default: "avail-light-client/rust-client")
-libp2p_identify_agent = "avail-light-client/rust-client"
-# Vector of Relay nodes, which are used for hole punching (default: empty)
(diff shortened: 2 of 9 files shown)


src/network/client.rs
```diff
@@ -145,29 +145,27 @@ impl Client {
...
 			if self
 				.sender
 				.send(Command::PutKadRecordBatch {
 					records,
 					quorum,
-					sender,
+					sender: tx.clone(),
 				})
 				.await
 				.context("Command receiver should not be dropped.")
```
aterentic-ethernal, line 158: Out of a scope of this change, but in this case error is discarded, so context is not needed. Also, in case of send error, num_success will always be 0, since send is blocking. This is clearer if code is split:

```

File: src/network/event_loop.rs [hunk 2]
```diff
@@ -188,57 +241,105 @@ impl EventLoop {
...
+									.and_then(|queries| {
+										let success_count = queries
 											.iter()
-											.filter(|(_, qs)| matches!(qs, QueryState::Succeeded))
+											.filter(|(_, qd)| {
+												matches!(qd.status, Some(QueryStatus::Succeeded))
+											})
 											.count();
+										// send result back
+										_ = ch.send(NumSuccPut(success_count));
```
aterentic-ethernal, line 327: Should we await for this?

src/network/event_loop.rs
```diff
@@ -248,17 +349,25 @@ impl EventLoop {
...
+									if let Some(QueryDetails {
+										res_sender: QueryChannel::Bootstrap(ch),
+										..
+									}) = self.pending_kad_queries.remove(&id)
 									{
 										_ = ch.send(Ok(()));
+										// we can say that the startup bootstrap is done here
+										if self.bootstrap.is_startup_done == false {
+											self.bootstrap.is_startup_done = true;
+										}
```
aterentic-ethernal, line 361: I guess its ok to remove condition:
```
									// we can say that the startup bootstrap is done here
										self.bootstrap.is_startup_done = true;
```

src/network/event_loop.rs
```diff
@@ -188,57 +241,105 @@ impl EventLoop {
...
-
-								// TODO: Handle or log errors
-								*v = match result {
-									Ok(_) => QueryState::Succeeded,
-									Err(error) => QueryState::Failed(error.into()),
-								};
-
-								let has_pending = self
-									.pending_kad_query_batch
+								}
```
aterentic-ethernal, line 286: wdyt of:
```
								// Set status for ongoing KAD query based on result from PUT
								// TODO: Handle or log errors
								qd.status = match result {
									Ok(record) => {
										// check if removal of local records is set for fat clients memory optimization
										if self.kad_remove_local_record {
											self.swarm
												.behaviour_mut()
												.kademlia
												.remove_record(&record.key);
										}
										Some(QueryStatus::Succeeded)
									},
									Err(err) => Some(QueryStatus::Failed(err.into())),
								}
```

aterentic-ethernal (approved):

src/network/client.rs
```diff
@@ -144,29 +144,29 @@ impl Client {
...
 
-			num_success +=
-				if let Ok(NumSuccPut(num)) = receiver.await.context("Sender not to be dropped.") {
-					num
-				} else {
-					num_success
-				};
+		// drop tx manually,
+		// ensure that only senders in spawned threads are still in use
+		drop(tx);
```
aterentic-ethernal, line 165: I guess if we move first reference into separate thread this won't be needed.

src/network/event_loop.rs
```diff
@@ -188,57 +242,107 @@ impl EventLoop {
...
+
+								// if there are no Pending queries left
+								// collect ones that are considered to be done
+								let finished_queries = has_no_pending
+									.then(|| {
+										batch_queries
+											.into_iter()
+											.filter(|(_, qd)| {
+												!matches!(qd.status, QueryStatus::Pending)
+											})
```
aterentic-ethernal, line 309: Maybe simpler `qd.status != QueryStatus::Pending`?

src/network/event_loop.rs
```diff
@@ -188,57 +242,107 @@ impl EventLoop {
...
+											.collect::<HashMap<&QueryId, &QueryDetails>>()
+									})
+									// count successful ones and gather what ids to remove
+									.and_then(|queries| {
+										let success_count = queries
 											.iter()
-											.filter(|(_, qs)| matches!(qs, QueryState::Succeeded))
+											.filter(|(_, &qd)| {
+												matches!(qd.status, QueryStatus::Succeeded)
+											})
```
aterentic-ethernal, line 318: Same here since there is no need for complex pattern match?

.gitignore
```diff
@@ -1,6 +1,5 @@
+.*
```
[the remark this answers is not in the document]
momosh-ethernal: wth, fixing immediately

README.md
```diff
@@ -81,29 +81,31 @@ http_server_port = "7000"
...
-secret_key = { key =  "1498b5467a63dffa2dc9d9e069caf075d16fc33fdd4c3b01bfadae6433767d93" }
-# Libp2p service port range (port, range) (default: 37000).
-libp2p_port = "37000"
-# Configures LibP2P TCP port reuse for local sockets, which implies reuse of listening ports for outgoing connections to enhance NAT traversal capabilities (default: false)
-libp2p_tcp_port_reuse = false
-# Configures LibP2P AutoNAT behaviour to reject probes as a server for clients that are observed at a non-global ip address (default: false)
-libp2p_autonat_only_global_ips = false
-# Libp2p AutoNat throttle period for re-using a peer as server for a dial-request. (default: 1 sec)
-libp2p_autonat_throttle = 1
+secret_key = { seed="1" }
```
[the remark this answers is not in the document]
momosh-ethernal: a leftover

PR #230, opened 2023-06-26
Added packet analyzer using pcap

sh3ll3x3c:
- created `network_analyzer` mod, hidden behind a feature flag
- `libpcap` used for packet sniffing on set libp2p port
- both `lo` and external NIC are used to listen
- sampling interval determines the serialization frequency which is a placeholder for now

### DIFF
--- a/src/network/network_analyzer.rs
+++ b/src/network/network_analyzer.rs
@@ -95,16 +95,10 @@ fn start_listening_on_device(
 	Ok(())
 }
 
-fn open_capture_from_device(device_name: String) -> Result<Capture<Active>> {
-	match pcap::Capture::from_device(device_name.as_str()) {
-		Ok(l_c) => match l_c.immediate_mode(true).promisc(true).open() {
-			Ok(l_c_o) => Ok(l_c_o),
-			Err(err) => {
-				return Err(err.into());
-			},
-		},
-		Err(err) => {
-			return Err(err.into());
-		},
-	}
+fn open_capture_from_device(device_name: String) -> Result<Capture<Active>, pcap::Error> {
+	let l_c = Capture::from_device(device_name.as_str())?
+		.immediate_mode(true)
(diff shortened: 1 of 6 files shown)


src/network/network_analyzer.rs
```diff
@@ -0,0 +1,86 @@
...
+		.filter(&format!("udp port {port}"), true)
+		.unwrap();
+
+	let mut dev_cap =
+		pcap::Capture::from_device(dev.clone().expect("device should exist").name.as_str())
+			.unwrap()
+			.immediate_mode(true)
+			.promisc(true)
+			.open()
+			.unwrap();
```
momosh-ethernal, line 50: All unwraps from lines 38 through here cause panic.
Do we want that?

src/network/network_analyzer.rs
```diff
@@ -0,0 +1,86 @@
...
+	dev_cap.filter(&format!("udp port {port}"), true).unwrap();
+
+	let total_bytes = Arc::new(AtomicU32::new(0));
+
+	// Start packet inspection tasks
+	let total_bytes_lo = Arc::clone(&total_bytes);
+	tokio::task::spawn_blocking(move || loop {
```
momosh-ethernal, line 57: Keep in mind that closures spawned with `spawn_blocking` can't be canceled abruptly; no low-level API exists that can stop the thread from running. And since we don't have any `cancelation tokens` in place, when we shut down the executor, it will wait indefinitely for all blocking operations to finish.

Maybe we should use that `shutdown_timeout` here?

momosh-ethernal (approved):

src/network/network_analyzer.rs
sh3ll3x3c: Absolutely not, slipped my mind.

momosh-ethernal (approved):

PR #234, opened 2023-06-29
Crawl DHT block.

aterentic-ethernal:


src/crawl_client.rs
```diff
@@ -0,0 +1,66 @@
...
+) {
+	info!("Starting crawl client...");
+
+	while let Ok((header, received_at)) = message_rx.recv().await {
+		if let Some(seconds) = delay_for(Duration::from_secs(delay), received_at.elapsed()) {
+			info!("Sleeping for {seconds:?} seconds");
+			tokio::time::sleep(seconds).await;
+		}
+		let block_number = header.number;
+		info!("Crawling block {block_number}...");
```
i1i1: `tracing` supports structural logging. It is the best to use field syntax. Like this:
```rust
info!(block_number, "Crawling block...");
```

More on this [here](https://docs.rs/tracing/latest/tracing/#recording-fields)

src/crawl_client.rs
```diff
@@ -0,0 +1,66 @@
...
+
+		let start = SystemTime::now();
+
+		let Ok(block) = types::BlockVerified::try_from(header) else {
+			error!("Header is not valid");
+			continue;
+		};
```
i1i1, line 42: We can report an error here

src/types.rs
```diff
@@ -310,7 +314,7 @@ pub struct LightClientConfig {
 	pub ttl: u64,
 }
 
-fn delay_for(target: Duration, elapsed: Duration) -> Option<Duration> {
+pub fn delay_for(target: Duration, elapsed: Duration) -> Option<Duration> {
 	(target > elapsed).then(|| target - elapsed)
 }
```
i1i1, line 321: I think `delay_for` is a weird function, as it literally is the same as [`std::time::Duration::checked_sub`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.checked_sub).

src/crawl_client.rs
i1i1: We can also move this check before the sleep, so we won't waste queue for not valid headers

src/types.rs
```diff
@@ -487,6 +489,14 @@ impl From<&RuntimeConfig> for AppClientConfig {
 	}
 }
 
+#[derive(Serialize, Deserialize, PartialEq, Clone, Debug)]
```
i1i1, line 492: `Copy` derive is missing. I think clippy should warn for this

i1i1: You can also derive Default here and mark `Cells` as default variant

src/crawl_client.rs
```diff
@@ -0,0 +1,66 @@
...
+
+	while let Ok((header, received_at)) = message_rx.recv().await {
+		if let Some(seconds) = delay_for(Duration::from_secs(delay), received_at.elapsed()) {
+			info!("Sleeping for {seconds:?} seconds");
+			tokio::time::sleep(seconds).await;
+		}
+		let block_number = header.number;
+		info!("Crawling block {block_number}...");
+
+		let start = SystemTime::now();
```
i1i1, line 37: `SystemTime` can move forward and backward as our system syncs time. It is the best to use `std::time::Instant` as it is monotonic and nondecreasing.

i1i1: Btw, that is why `elapsed` on `SystemTime` returns `Result` and not just `Duration`

src/crawl_client.rs
```diff
@@ -40,27 +41,51 @@ pub async fn run(
 			continue;
 		};
 
-		let positions = block
-			.dimensions
-			.iter_extended_partition_positions(&ENTIRE_BLOCK)
-			.collect::<Vec<_>>();
+		if mode == CrawlMode::Cells || mode == CrawlMode::Both {
```
i1i1, line 44: A bit shorter version `if matches!(mode, CrawlMode::Cells | CrawlMode::Both) {`

i1i1 (commented):
Apart from those comments, looks good to me!

I can't dig very deep into logic yet, but hope it is not overwhelming and somewhat relevant.

src/crawl_client.rs
aterentic-ethernal: Precisely, 1 character shorter :D

src/types.rs
aterentic-ethernal: There is no clippy warning, for what is `Copy` derive needed?

i1i1: Okay, there is a [rustc lint](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.MISSING_COPY_IMPLEMENTATIONS.html) for that. We can enable it.

Copy is needed for declaring that some type can be fully allocated on stack. That enables a bunch of optimizations (basically copying `Copy` type will just use `memcpy` and deallocation of this thing on stack can also be skipped)

src/crawl_client.rs
i1i1: But is way better for reading :D

src/crawl_client.rs
aterentic-ethernal: I have moved it before the sleep, but there is no significant change since sleep is used to postpone each block processing for at least given amount of second, and each block is produced in regular intervals (e.g. 20 seconds).

src/crawl_client.rs
```diff
@@ -0,0 +1,90 @@
...
+			let success_rate = fetched.len() as f64 / total as f64;
+			info!(
+				block_number,
+				"Fetched {fetched} cells of {total}, success rate: {success_rate}",
+				fetched = fetched.len(),
+			);
+			let _ = metrics.record(MetricValue::CrawlCellsSuccessRate(success_rate));
+		}
+
+		if mode == CrawlMode::Rows || mode == CrawlMode::Both {
```
i1i1, line 67: Same for here
```suggestion
		if matches!(mode, CrawlMode::Rows | CrawlMode::Both) {
```

src/types.rs
```diff
@@ -481,6 +487,14 @@ impl From<&RuntimeConfig> for AppClientConfig {
 	}
 }
 
+#[derive(Serialize, Deserialize, PartialEq, Clone, Copy, Debug)]
+#[serde(rename_all = "kebab-case")]
+pub enum CrawlMode {
+	Rows,
+	Cells,
+	Both,
+}
```
i1i1, line 496:
```suggestion
#[derive(Serialize, Deserialize, PartialEq, Clone, Copy, Debug, Default)]
#[serde(rename_all = "kebab-case")]
pub enum CrawlMode {
	Rows,
	#[default]
	Cells,
	Both,
}
```

src/subscriptions.rs
```diff
@@ -257,15 +259,19 @@ async fn subscribe_check_and_process(
 						},
 					};
 
-					message_tx.send((header, received_at)).await?;
+					if let Err(error) = message_tx.send((header, received_at)) {
+						error!("Sending header failed: {error}");
```
i1i1, line 263:
```suggestion
						error!(%error, "Sending header failed");
```

src/subscriptions.rs
```diff
@@ -257,15 +259,19 @@ async fn subscribe_check_and_process(
...
 				}
 
 				info!("Sending finalized block {}", header.number);
 				// Reset last finalized block
 				last_finalized_block_header = header.clone();
 
 				// Finally, send the verified block (header)
-				message_tx.send((header, received_at)).await?;
+				if let Err(error) = message_tx.send((header, received_at)) {
+					error!("Sending verified header failed: {error}");
```
i1i1, line 273:
```suggestion
					error!(%error, "Sending verified header failed");
```

src/crawl_client.rs
```diff
@@ -0,0 +1,90 @@
...
+				.iter()
+				.step_by(2)
+				.flatten()
+				.count();
+
+			let success_rate = fetched as f64 / total as f64;
+			info!(
+				block_number,
+				"Fetched {fetched} rows of {total}, success rate: {success_rate}"
+			);
```
i1i1, line 83:
```suggestion
			info!(block_number, success_rate, total_rows = total, "Fetched {fetched} rows");
```

src/crawl_client.rs
momosh-ethernal: And much harder for debugging

sh3ll3x3c (approved):

PR #241, opened 2023-07-05
bump versions w.r.t runtime upgrade

kroos47:


src/main.rs
```diff
@@ -247,8 +247,8 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 	let last_full_node_ws = data::get_last_full_node_ws_from_db(db.clone())?;
 
 	let version = rpc::Version {
-		version: "1.6.1".to_string(),
-		spec_version: 10,
+		version: "1.6.1-7fcca43d5f5".to_string(),
```
0xSasaPrsic, line 250: I don't think that hash is necessary.

prabal-banerjee: I agree. This might mess up the version output.

prabal-banerjee (approved):

PR #243, opened 2023-07-25
Update `avail-core` dep

fmiguelgarcia:
- Add new shared/common `avail-core` dep.
- Uses `dprint` to format `toml` and `md` files.

## Companion PRs
- [Avail Core PR 27](https://github.com/availproject/avail-core/pull/27)
- [Avail Node PR 183](https://github.com/availproject/avail/pull/183)


Cargo.toml
```diff
@@ -1,72 +1,68 @@
...
+edition = "2021"
+license = "Apache-2.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-kate-recovery = { version = "0.8", git = "https://github.com/availproject/avail-core", tag = "da-primitives/v0.4.6" }
-avail-subxt = { git = "https://github.com/availproject/avail.git", tag = "v1.6.1-rc3" }
+# Internal deps
+avail-core = { path = "../avail-core/core" }
```
kroos47, line 12: we should change the path after merging the avail-core

aterentic-ethernal (changes requested):
All good, things are cleaner in some cases now 👍 I guess we have to remove licence from this PR since we have to check dependencies and we don't want to open source it yet. Also can we move version check change to separate PR and remove spec_version check there also?

aterentic-ethernal:
Also, it would be great to split dprint changes into separate PR to be able to merge it before anything else, and avoid conflicts later.

fmiguelgarcia:
> Also, it would be great to split dprint changes into separate PR to be able to merge it before anything else, and avoid conflicts later.

Dprint changes have been moved to https://github.com/availproject/avail-light/pull/249
att. @aterentic-ethernal

kroos47 (approved):

src/utils.rs
```diff
@@ -40,11 +43,21 @@ pub(crate) fn extract_kate(extension: &HeaderExtension) -> (u16, u16, H256, Vec<
...
+
+	let size = compact.size;
+	let index = compact
+		.index
+		.iter()
+		.map(|item| DataLookupItem::new(AppId(item.app_id.0), item.start))
+		.collect::<Vec<_>>();
+
+	let compact = CompactDataLookup::new(size, index);
+	DataLookup::try_from(compact).expect("Invalid app lookup")
```
aterentic-ethernal, line 60: We can return error instead of panic here, since this is used in `process_block` and `From` implementation for `BlockVerified` type, which both supports returning error.

aterentic-ethernal (approved):
All good from my side, except that expect (:D). I guess this will be squash merge so we could loose reasoning behind version and metadata check change, so if possible, we can do that in two commits or add a comment in the code.

PR #250, opened 2023-08-04
A simple tool to test if Light Client is compatible with a particular version of Avail node

luka-ethernal:


Cargo.lock
```diff
@@ -669,8 +669,8 @@ dependencies = [
  "serde",
  "serde_json",
  "smallvec 1.11.0",
- "sp-arithmetic 16.0.0",
- "sp-core 21.0.0",
+ "sp-arithmetic 6.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
```
fmiguelgarcia, line 672: This is really weird change, could you pls regenerate `Cargo.lock` and double check it?

momosh-ethernal (approved):

Cargo.lock
luka-ethernal: yes, thanks, didn't notice it, I didn't want to change anything in Cargo.lock at all!

aterentic-ethernal (approved):

PR #264, opened 2023-08-18
Add SyncFinality as a new client/task.

luka-ethernal:
Todo:

- ~~Remove finality checks currently in the sync client~~
- ~~Save the verified block height in the db and resume from there (both in sync_finality as well as subscription).~~

### DIFF
--- a/src/subscriptions.rs
+++ b/src/subscriptions.rs
@@ -161,6 +161,7 @@ async fn subscribe_check_and_process(
 			},
 		}
 
+		let mut finality_synced = false;
 		while let Some(justification) = justifications.pop() {
 			// Iterate through headers and try to find a matching one.
 			if let Some(pos) = unverified_headers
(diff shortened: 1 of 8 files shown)


luka-ethernal:
This PR finally implements the complete finalization check and synchronization. This is how it works:

- At LC startup, it is checked whether we have a checkpoint stored and if we do, we move on from there. Otherwise, we start from genesis.
- It is not possible to check finality of an arbitrary block in the past. Rather, it is only possible to check blocks that have validator set changes. These blocks then cover all the previous ones until the block that also had validator set changes (which is how GRANDPA works).
- We check finality only on those blocks, and just hash and check parent hash for all other blocks. This is enough to guarantee finality has been achieved. We only save checkpoints on blocks where we directly check finality.
- When the finality is checked up until the head block, we mark the finality as synced and switch to listening for justifications, which usually happens for every block under normal circumstances (no dropped validators etc). We also save checkpoints, provided the sync has been completed.

There are a lot of details missing from this explanation, such as difficulties of obtaining validator set at genesis (and corresponding GRANDPA public keys), if anyone wants, I can explain further.

src/sync_finality.rs
```diff
@@ -0,0 +1,341 @@
...
+		D: serde::Deserializer<'de>,
+	{
+		let data = from_hex(&String::deserialize(deserializer)?)
+			.map_err(|e| de::Error::custom(format!("{:?}", e)))?;
+		Decode::decode(&mut &data[..]).map_err(|e| de::Error::custom(format!("{:?}", e)))
+	}
+}
+
+#[async_trait]
+#[automock]
```
aterentic-ethernal, line 69: Do we need the mock? If we want to test the flow, we would need to abstract db an client, and this trait would have a bunch of functions.

src/subscriptions.rs
```diff
@@ -257,6 +216,19 @@ async fn subscribe_check_and_process(
 					break 'mainloop Err(anyhow!(
 						"Not signed by the supermajority of the validator set."
 					));
+				} else {
+					let state_locked = state.lock().unwrap();
+					if state_locked.finality_synced {
```
aterentic-ethernal, line 221: Do we need lock here or we can afford skipping one checkpoint sometimes by:
```
let finality_sync = state.lock().unwrap().finality_sync;
if finality_sync {
```

aterentic-ethernal (approved):

PR #267, opened 2023-08-22
V2 API: Add submit endpoint to V2 API

aterentic-ethernal:


src/api/v2/transactions.rs
```diff
@@ -0,0 +1,89 @@
...
+	async fn submit(&self, data: Vec<u8>) -> Result<H256>;
+}
+
+#[derive(Clone)]
+pub struct RawSubmit {
+	pub node_client: avail::Client,
+	pub app_id: u32,
+}
+
+// NOTE: Replace this with avail::PairSigner after implenting required traits in avail-subxt
```
i1i1, line 23: I think `TODO` should be more relevant here

src/http.rs
i1i1, line 1: This file is empty :(

src/api/v2/transactions.rs
```diff
@@ -0,0 +1,89 @@
...
+			node_client,
+			app_id,
+		}),
+		Some(pair_signer) => Submitter::SignAndSubmit(Box::new(SignAndSubmit {
+			node_client,
+			app_id,
+			pair_signer,
+		})),
+	}
+}
```
i1i1, line 89: Is it a constructor for `Submitter`?

i1i1 (commented):
Very minor comments. Overall LGTM!

src/api/v2/transactions.rs
aterentic-ethernal: Makes sense 👍

src/http.rs
aterentic-ethernal: Not sure how that ended here O.o Deleted.

src/api/v2/transactions.rs
aterentic-ethernal: You are right, but it was removed anyway :)

src/api/v2/types.rs
```diff
@@ -80,6 +81,24 @@ pub struct Status {
...
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum Transaction {
+	Data(String),
+	Extrinsic(String),
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct TransactionHash {
+	pub hash: String,
```
i1i1, line 93: You can use `sp_core::Hash` directly here. It implements serialize and that way it will be deserialized smth like this:
```
{"hash":"0x0000000000000000000000000000000000000000000000000000000000000000"}
```

i1i1: I would personally just remove this structure at all and remove just `warp::reply::Json<sp_core::H256>` directly from `submit` method

src/types.rs
```diff
@@ -292,6 +292,8 @@ pub struct RuntimeConfig {
 	pub max_kad_record_size: u64,
 	/// The maximum number of provider records for which the local node is the provider. (default: 1024).
 	pub max_kad_provided_keys: u64,
+	/// Avail account secret key. (default: None)
+	pub avail_secret_key: Option<String>,
```
i1i1, line 296: I think you can create a wrapper for key pair which will deserialize it from string which can be reused later

src/api/v2/handlers.rs
```diff
@@ -26,6 +31,33 @@ pub async fn subscriptions(
...
 
+pub async fn submit(
+	submitter: Arc<impl transactions::Submit>,
+	transaction: Transaction,
+) -> Result<TransactionHash, Error> {
+	fn base64_decode(data: String) -> Result<Vec<u8>, Error> {
+		general_purpose::STANDARD
+			.decode(data)
+			.map_err(|error| Error::bad_request(format!("Bad Request: {error}")))
+	}
```
i1i1, line 42: I think it is better to create a type wrapper for base64 decoding. Like:
```rust
#[derive(Deserialize)]
#[serde(try_from = "String")]
pub struct Base64(pub Vec<u8>);

impl TryFrom<String> for Base64 { ... }
```

src/api/v2/handlers.rs
```diff
@@ -26,6 +31,33 @@ pub async fn subscriptions(
...
+
+	if matches!(&transaction, Transaction::Data(_)) && !submitter.has_signer() {
+		return Err(Error::not_found());
+	};
+
+	let transaction = match transaction {
+		Transaction::Data(data) => transactions::Transaction::Data(base64_decode(data)?),
+		Transaction::Extrinsic(data) => transactions::Transaction::Extrinsic(base64_decode(data)?),
+	};
```
i1i1, line 51: Also there are 2 types of transactions here which is weird. You can squash them into one type and make it generic over container inside, or you can just have separate enum like `TransactionType`.

src/api/v2/types.rs
aterentic-ethernal: Still unsure if we want `0x` prefix or not, but Ill leave it with prefix as it si simpler atm, and get back to this later. At this point there will be more fields in respose (which can change in the future), so second comment doesn't apply.

src/api/v2/handlers.rs
aterentic-ethernal: This is cool, but we want to have custom bad request message in the response if deserialization fails, and instead of handling custom error in rejections it seems there is no other way (but I have to look more on it). That would resolve both having double structure for transaction and make things simpler by avoiding using generics.

src/api/v2/handlers.rs
aterentic-ethernal: Separate enum can imply different interface (e.g. `{"type":"da","data":"asdfaa88sfa"}` instaed, which is something Ill consider.

src/api/v2/handlers.rs
i1i1: You can customize Error type for that purpose. I didn't get where are generics here also

src/api/v2/mod.rs
```diff
@@ -217,6 +258,47 @@ mod tests {
...
+	async fn submit_route_bad_request() {
+		let route = super::submit_route(Some(Arc::new(MockSubmitter {})));
+		let response = warp::test::request()
+			.method("POST")
+			.path("/v2/submit")
+			.body(r#"{"data":"dHJhbnooNhY3Rpb24:"}"#)
+			.reply(&route)
+			.await;
+		assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+		assert_eq!(response.body(), "Bad Request: Invalid byte 58, offset 17.");
```
i1i1, line 289: I think there is no need to check for exact error message as it might change and break test. Checking for status code sounds like enough.

src/api/v2/types.rs
i1i1: I think it doesn't matter if there will be a prefix. So for ease lets keep it.

As for second comment. The structure is called `TransactionHash` which is really weird, so if we have plans to add fields lets call it `SubmitResponse` or smth

src/api/v2/handlers.rs
i1i1: We are not decided for an interface, so we can switch it right now with ease. So I think `TransactionType` sounds way better than having 2 same types

src/api/v2/types.rs
i1i1: Sorry, didn't saw last force push. All good!

src/api/v2/mod.rs
aterentic-ethernal: Not sure about that one, we should have informative messages to lower confusion of API users during development, which should be part of a test. This is of course just error from json deserialization and could be fragile, but it might be good enough.

src/api/v2/handlers.rs
aterentic-ethernal: Its not that easy because warp has default string prefix, which is maybe ok, Ill keep it simple atm and later resolve this issue if needed.

src/api/v2/handlers.rs
aterentic-ethernal: This is fine at the moment since there are a lot unknowns, so lets keep it like this and continue discussion on the https://github.com/availproject/RFCs/discussions/3.

src/api/v2/mod.rs
aterentic-ethernal: Lets keep this one also and decide in the future, since there will be more tests and it will be clearer what they represent (also this is still draft PR lol)

src/api/v2/handlers.rs
i1i1: RFC doesn't feature `extrinsic` for `/submit` though. For that, body will look like `{ "extrinsic": "adf1234" }`.

Let's move the discussion there, but I think RFC is not up-to-date if you've decided to have extrinsic variant also

src/api/v2/mod.rs
i1i1: We will have informative messages, but there is no way to check reliably for it. Imagine you update `serde_json` and message format will change. You would need to change string again, and it is literally a waste of time which won't give you any benefit.

src/api/v2/transactions.rs
```diff
@@ -16,6 +18,12 @@ pub trait Submit {
 // TODO: Replace this with avail::PairSigner after implementing required traits in avail-subxt
 pub type AvailSigner = PairSigner<AvailConfig, Pair>;
 
+impl From<&AvailSecretKey> for AvailSigner {
+	fn from(value: &AvailSecretKey) -> Self {
+		AvailSigner::new(value.0.clone())
+	}
+}
```
i1i1, line 25: It is best to push cloning to user side, so that he could control allocations.
```suggestion
impl From<AvailSecretKey> for AvailSigner {
	fn from(value: AvailSecretKey) -> Self {
		AvailSigner::new(value.0)
	}
}
```

src/api/v2/handlers.rs
aterentic-ethernal: `Bad request: {error}` was custom message, now warp is giving `Request body deserialize error: {error}` since more general solution was implemented. I guess both are fine, and I'm ok with setling on this at the moment, but latter we need to think how end users interpret those messages and try to keep consistency. From my perspective, its ok to crash since we don't loose track that something was actually changed. If it bothers, it can be easily be removed.

src/types.rs
```diff
@@ -293,7 +295,38 @@ pub struct RuntimeConfig {
...
+	}
+}
+
+impl std::fmt::Debug for AvailSecretKey {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		f.debug_tuple("AvailSecretKey")
+			.field(&self.0.to_raw_vec())
+			.finish()
+	}
+}
```
i1i1, line 320: I think it is a bad idea to implement Debug for it. The best would be to skip this field in other structures. There is a reason why substrate guys didn't derive Debug for secret key, righ.

aterentic-ethernal: Sure, I have to check it once how to avoid it, but didn't wanted to left uncommited.

src/api/v2/handlers.rs
i1i1: I've made a thread for RFC for discussion of the API: https://github.com/availproject/RFCs/discussions/3#discussioncomment-6876407

src/api/v2/mod.rs
i1i1: IKR, but my point is that we don't have to test very very very specific behavior, like how the error is printed.

src/api/v2/handlers.rs
i1i1: I mean there is status code already, no need to duplicate it, right? So we can rely on defaults.

aterentic-ethernal: Status code 400 can imply both json parsing and base 64 decoding issue, so message is what make them distinct.

src/api/v2/transactions.rs
i1i1: What do you mean? `subxt::ext::sp_core::sr25519::Pair` is the same as `AvailSecretKey` as it just wraps `Pair`. Cloning will be the same in the end.

src/types.rs
aterentic-ethernal: So, I have removed field from serialization since default is none anyway, and for debugging, i have left public key which could be useful for identification.

src/types.rs
```diff
@@ -292,6 +294,31 @@ pub struct RuntimeConfig {
 	pub max_kad_record_size: u64,
 	/// The maximum number of provider records for which the local node is the provider. (default: 1024).
 	pub max_kad_provided_keys: u64,
+	/// Avail account secret key. (default: None)
+	#[serde(skip)]
```
i1i1, line 298: I think it will skip it also during deserialization: https://serde.rs/field-attrs.html#skip

src/api/v2/handlers.rs
i1i1: Sorry, looked at the code again. LGTM right now! Thanks a lot! :smiley:

src/api/v2/mod.rs
aterentic-ethernal: Understand, but its important what we communicate to end users in case of different errors, so let just think that way here.

i1i1: Of course, but test != communication. Test is related to testing rather. If you want specific error messages, you can document them in RFC or in docs.

aterentic: I would say that tests actually communicate something, like presence of the bugs. But thats not the point here, point is that I want to confirm that two different inputs gives two different error messages with the same status, and thats impossible without checking error message somehow, what's the thing I'm missing here?

aterentic-ethernal: would say that tests actually communicate something, like presence of the bugs. But thats not the point here, point is that I want to confirm that two different inputs gives two different error messages with the same status, and thats impossible without checking error message somehow, what's the thing I'm missing here?

src/types.rs
aterentic-ethernal: Right, fixed 👍

src/api/v2/mod.rs
i1i1: Of course, they communicate something, but that is the last thing which does it:
1. Actual API which literally dictates you what to do
2. Docs for that API which cover some edge cases and spill the light on some corner cases
3. And only right here are tests

My point is that there is no need to check for exact error message. If it was some kind of library, you could have checked for exact error variant, while discarding some unnecessary stuff. Like:
```rust
enum Error {
  A,
  B,
  C(std::io::Error)
}

fn f(x: i32) -> Result<(), Error>;

#[test]
fn test_f() {
	assert_matches!(f(0), Ok(()));
	assert_matches!(f(-1), Err(C(_)));
}
```

But here you just have literal string, which content still has possibility to change with some version of this or that library. There is no good way to check that this error meaning didn't change. So let's just not do it and check for error code.

aterentic-ethernal: Right way of implementing it is not exposing error messages from any library to end users. but this is a trade off which needs to be at least controlled somehow.

i1i1: Why not to expose error messages, though? Clearly you can't own everything and clearly `serde_json` know better what is the problem with user's json. My point is that there is no need to test serde, especially in this way

i1i1: Anyway, it looks a bit better now. It is not a critical thing, so LGTM!

src/api/v2/types.rs
```diff
@@ -80,6 +84,50 @@ pub struct Status {
 	pub partition: Option<Partition>,
 }
 
+#[derive(Deserialize)]
```
i1i1, line 87:
```suggestion
#[derive(Debug, Clone, Deserialize)]
```

src/api/v2/types.rs
```diff
@@ -80,6 +84,50 @@ pub struct Status {
...
+	Data(Base64),
+	Extrinsic(Base64),
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SubmitResponse {
+	pub block_hash: H256,
+	pub hash: H256,
+	pub index: u32,
+}
```
i1i1, line 123: Same derives here. I think it is best to add a lint, which would check that.

src/api/v2/transactions.rs
```diff
@@ -0,0 +1,69 @@
...
+// TODO: Replace this with avail::PairSigner after implementing required traits in avail-subxt
+pub type AvailSigner = PairSigner<AvailConfig, Pair>;
+
+impl From<AvailSecretKey> for AvailSigner {
+	fn from(value: AvailSecretKey) -> Self {
+		AvailSigner::new(value.0)
+	}
+}
+
+#[derive(Clone)]
```
i1i1, line 27:
```suggestion
#[derive(Clone, Debug)]
```

i1i1 (commented):
LGTM, Apart from duplication of Transaction types!

aterentic-ethernal:
> LGTM, Apart from duplication of Transaction types!

Which duplication, I have removed duplicated types by introducing Base64 wrapper type?

luka-ethernal (approved):

i1i1 (approved):

PR #271, opened 2023-08-29
Add nix flake for dev shell and for packaging

i1i1:
This pr adds nix flake for reusable and reproducible dev environment. Flake is derived from an example from [here](https://crane.dev/examples/quick-start.html).

Nix is a package manager for reproducible builds and dev environment[(1)]. Flakes are an unstable feature, which is used by everyone, and which pushes reproducibility of environments even further[(2)].

[(1)]: https://nixos.wiki/wiki/Nix_package_manager
[(2)]: https://nixos.wiki/wiki/Flakes

If noone minds, I would prefer at least to have a simple version of it which would reproduce simple dev environment without packaging the avail light client itself (it would remove like 100 lines of code).

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .DS_Store
 config.yaml
 avail_light_store
+/.direnv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,150 @@
+{
+  "nodes": {
+    "crane": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "flake-utils": [
+          "flake-utils"
+        ],
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-overlay": "rust-overlay"
+      },
+      "locked": {
+        "lastModified": 1693163878,
+        "narHash": "sha256-HXuyMUVaRSoIA602jfFuYGXt6AMZ+WUxuvLq8iJmYTA=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "43db881168bc65b568d36ceb614a0fc8b276191b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": []
+      },
+      "locked": {
+        "lastModified": 1693203690,
+        "narHash": "sha256-qrgyFtRaduofcJF1T7TscNWM4HTx4qtFzpilD0K8M0o=",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "8eb35c8c402edda73a7dde82e48bf279542640c7",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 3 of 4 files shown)


luka-ethernal:
I think this repo isn't a place for it. Being in the repo means it's officially supported, which means we need to maintain it. None of us use NixOS and, while it is cool, I think it's too big of a niche to officially support it. It can live in a separate repo, though. Also, when I have files/scripts specific to my workflow, I just add them to .gitignore or just leave them untracked.

i1i1:
> Being in the repo means it's officially supported, which means we need to maintain it. ... None of us use NixOS 

I personally use it, so you can count that even in terms of maintaining it.

> I think it's too big of a niche to officially support it

If it is too niche, then it is not a big deal to break support for it in the future. If noone will be able to support it, we can just remove it, not big of a deal.

momosh-ethernal:
> I personally use it, so you can count that even in terms of maintaining it.

That is so stunning! But, please continue using it without being tracked, on your local env.

> If it is too niche, then it is not a big deal to break support for it in the future. If noone will be able to support it, we can just remove it, not big of a deal.

It is a pigeonhole. There is no sense in adding something that is going to be removed shortly.
This contaminates the source and is unnecessary.
If project team members agree, this could live in a separate repo under the organization.

sh3ll3x3c:
Agreed. Let us not make these kinds of decisions in a hurry, bringing additional complexity without a clear tradeoff is not something we need right now.

i1i1:
> That is so stunning! But, please continue using it without being tracked, on your local env.

Didn't get what you mean here.

> It is a pigeonhole. There is no sense in adding something that is going to be removed shortly afterwards.

I'm not saying that it will be removed shortly, my take here is that I want to maintain it. If this thing is too niche then we can always remove it, and it is not big of a commitment, so we can just merge it and forget about it.

momosh-ethernal:
> Didn't get what you mean here.

This is a great mental exercise. A pretty nice example of what can be done, but completely unnecessary in this use case.
If you need it, and it helps you, please use it locally, in your own local development environment.

> ... so we can just merge it and forget about it

As mentioned above, we don't need contamination of this proportion.
Keep private tooling and helpers in local env, without being tracked.

i1i1:
> This is a great mental exercise. A pretty nice example of what can be done, but completely unnecessary in this use case.
If you need it, and it helps you, please use it locally, in your own local development environment.

Yes, but my point is that it might be reused the same way as `Dockerfile` is reused by someone else.

> Keep private tooling and helpers in local env, without being tracked.

I know, right, but it is not vs code settings extensions or vim settings. Though, something like that is already in repo like this: https://github.com/availproject/avail-light/blob/8e3a9a3888407f25fb2a14388cca9ba5c749628a/Dockerfile#L1

I didn't see how it contaminates repo with local environment or getting someone tracked, lol.

momosh-ethernal:
> Yes, but my point is that it might be reused the same way as Dockerfile is reused by someone else.

No one dockerizes anything unnecessarily and without prior need. You're right about Docker, it is being reused, across all levels of project organization.

> I didn't see how it contaminates repo with local environment or getting someone tracked, lol.

No need to be rude about it. No place nor context for this kind of comment.
The tracking remark was meant for the changes you have introduced to your local repo.

momosh-ethernal:
As commented, allowing the Nix ecosystem into the public repo would be considered and treated as a local environment contamination. 
Closing

i1i1:
Let's discuss it on dev sync

PR #278, opened 2023-09-01
Implement connection retry

i1i1:
Fixes https://github.com/availproject/engineering/issues/246

This pr adds a wrapper around avail client.
- A huge chunk of code is sitting in the first commit. It fixes implements reconnect and uses it everywhere apart from sync finality module and apart from subscriptions.
- Second commit fixes api compatibility binary
- 3rd one fixes sync finality to use our new backoff api
- and the last one implements reconnection streams

~I know that callsite api looks ugly right now, but there is no way around it apart from creating wrappers for those kinds of things. If you want to, I can go through each individual call to `rpc.with_client` and remove it.~

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -9,6 +9,7 @@ use avail_subxt::{
 	primitives::Header as DaHeader,
 	rpc::{types::BlockNumber, RpcParams},
 	utils::H256,
+	AvailConfig,
 };
 use futures::prelude::*;
 use itertools::Either;
@@ -24,6 +25,7 @@ use tokio::sync::RwLock;
 use tracing::{debug, info, instrument, warn};
 
 use crate::sync_finality::WrappedProof;
+use crate::types::AvailSecretKey;
 use crate::{consts::EXPECTED_NETWORK_VERSION, types::*};
 
 #[derive(Debug, Clone)]
@@ -33,6 +35,15 @@ pub struct RpcClient {
 	expected_version: ExpectedVersion<'static>,
 	db: Option<Arc<DB>>,
 	backoff: backoff::ExponentialBackoff,
+}
+
+// TODO: Replace this with avail::PairSigner after implementing required traits in avail-subxt
+pub type AvailSigner = subxt::tx::PairSigner<AvailConfig, subxt::ext::sp_core::sr25519::Pair>;
+
+impl From<AvailSecretKey> for AvailSigner {
+	fn from(value: AvailSecretKey) -> Self {
+		AvailSigner::new(value.0)
+	}
 }
 
 impl RpcClient {
@@ -516,6 +527,43 @@ impl RpcClient {
 		})
 		.await
 		.context("Version couldn't be retrieved, error")
+	}
+
+	pub async fn ext_sign_and_submit_then_wait_for_finalized_success<Call: subxt::tx::TxPayload>(
+		&self,
+		call: &Call,
+		signer: &AvailSigner,
+		other_params: avail_subxt::primitives::AvailExtrinsicParams,
+	) -> Result<subxt::blocks::ExtrinsicEvents<AvailConfig>> {
+		self.with_client(|client| {
+			let other_params = other_params.clone();
+			async move {
+				client
+					.tx()
+					.sign_and_submit_then_watch(call, signer, other_params)
+					.await?
+					.wait_for_finalized_success()
+					.await
+			}
+		})
+		.await
+	}
+
+	pub async fn ext_raw_submit_then_wait_for_finalized_success(
+		&self,
+		tx_bytes: Vec<u8>,
+	) -> Result<subxt::blocks::ExtrinsicEvents<AvailConfig>> {
+		self.with_client(|client| {
+			let tx_bytes = tx_bytes.clone();
+			async move {
+				subxt::tx::SubmittableExtrinsic::from_bytes(client, tx_bytes)
+					.submit_and_watch()
+					.await?
+					.wait_for_finalized_success()
+					.await
+			}
+		})
+		.await
 	}
 }
 
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -36,6 +36,7 @@ impl RpcClient {
 	async fn connect(
 		full_node_ws: &str,
 		expected_version: &ExpectedVersion<'_>,
+		expected_genesis_hash: Option<H256>,
 	) -> Result<(avail::Client, Node)> {
 		tracing::debug!(full_node_ws, "Trying to connect to rpc");
 		let client = build_client(&full_node_ws, false).await?;
@@ -56,7 +57,18 @@ impl RpcClient {
 		);
 
 		if !expected_version.matches(&system_version, &runtime_version.spec_name) {
-			return Err(anyhow!("expected {expected_version}, found {version}"));
+			return Err(anyhow!(
+				"expected version {expected_version}, found {version}"
+			));
+		}
+		let genesis_hash = client.genesis_hash();
+		match expected_genesis_hash {
+			Some(hash) if hash != genesis_hash => {
+				return Err(anyhow!(
+					"expected genesis hash {hash}, found {genesis_hash}"
+				));
+			},
+			_ => (),
 		}
 
 		info!("Connection established to the node: {full_node_ws} <{version}>");
@@ -64,7 +76,7 @@ impl RpcClient {
 			host: full_node_ws.to_owned(),
 			system_version,
 			spec_version: client.runtime_version().spec_version,
-			genesis_hash: client.genesis_hash(),
+			genesis_hash,
 		};
 
 		Ok((client, node))
@@ -73,11 +85,12 @@ impl RpcClient {
 	async fn connect_to_available_rpc(
 		full_nodes: &[String],
 		expected_version: &ExpectedVersion<'static>,
+		expected_genesis_hash: Option<H256>,
 	) -> Result<(avail::Client, Node)> {
 		full_nodes
 			.iter()
 			.map(|address| {
-				Self::connect(address, expected_version)
+				Self::connect(address, expected_version, expected_genesis_hash)
 					.inspect_err(move |error| warn!(address, %error, "Skipping connection"))
 			})
 			.collect::<futures::stream::FuturesUnordered<_>>()
@@ -114,10 +127,24 @@ impl RpcClient {
 			.transpose()?
 			.flatten();
 		Self::shuffle_full_nodes(&mut nodes, last_full_node.as_ref());
-		let (client, node) = Self::connect_to_available_rpc(&nodes, &expected_version).await?;
+		let expected_genesis_hash = if let Some(db) = &db {
+			crate::data::get_genesis_hash(db.clone())?
+		} else {
+			None
+		};
+		let (client, node) =
+			Self::connect_to_available_rpc(&nodes, &expected_version, expected_genesis_hash)
+				.await?;
+
+		info!(?node.genesis_hash);
 		if let Some(db) = &db {
 			crate::data::store_last_full_node_ws_in_db(db.clone(), node.host.clone())?;
+			if expected_genesis_hash.is_none() {
+				info!("No genesis hash is found in the db, storing the new hash now.");
+				crate::data::store_genesis_hash(db.clone(), node.genesis_hash)?;
+			}
 		}
+
 		let backoff = backoff::ExponentialBackoffBuilder::new()
 			.with_max_elapsed_time(Some(std::time::Duration::from_secs(20)))
 			.build();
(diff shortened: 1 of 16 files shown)


luka-ethernal:
As discussed, it would be prefferable if nodes rotation are in order of appearance in the config, instead of completely random, but that's not a big deal. Also, checking the genesis_hash of the new node is a must (you might have already added that).

src/rpc.rs
```diff
@@ -1,103 +1,378 @@
...
+		if let Some(db) = &db {
+			crate::data::store_last_full_node_ws_in_db(db.clone(), node.host.clone())?;
+			if expected_genesis_hash.is_none() {
+				info!("No genesis hash is found in the db, storing the new hash now.");
+				crate::data::store_genesis_hash(db.clone(), node.genesis_hash)?;
+			}
+		}
+
+		let backoff = backoff::ExponentialBackoffBuilder::new()
+			.with_max_elapsed_time(Some(std::time::Duration::from_secs(20)))
```
luka-ethernal, line 142: Perhaps extract the magic value as a const or even a config parameter for easier testing/tweaking?

src/rpc.rs
```diff
@@ -1,103 +1,378 @@
...
+		let (client, node) =
+			Self::connect_to_available_rpc(&nodes, &expected_version, expected_genesis_hash)
+				.await?;
+
+		info!(?node.genesis_hash);
+		if let Some(db) = &db {
+			crate::data::store_last_full_node_ws_in_db(db.clone(), node.host.clone())?;
+			if expected_genesis_hash.is_none() {
+				info!("No genesis hash is found in the db, storing the new hash now.");
+				crate::data::store_genesis_hash(db.clone(), node.genesis_hash)?;
```
luka-ethernal, line 138: Can we decouple the genesis hash check from RPC module? Move the reading from db to main and just pass Option<genesis_hash> to new(). Then also, if it was none, main can store the new genesis hash back to db.

i1i1: That would make api worse by either asking user to carry around db and pass it on each rpc call or just let him store things manually. I would prefer to keep it as it is

src/rpc.rs
```diff
@@ -1,103 +1,378 @@
...
+				warn!(%error, "Failed to connect to node. Trying to reach to another one");
+			},
+		}
+		let last_node = self.current_node().await;
+
+		let (ok, client) = backoff::future::retry(self.backoff.clone(), || async {
+			let mut f = f;
+			let mut nodes = self.nodes.clone();
+			Self::shuffle_full_nodes(&mut nodes, Some(&last_node.host));
+			let (client, node) = Self::connect_to_available_rpc(
```
luka-ethernal, line 176: Are we creating a different connection each time? What happens if we already have a valid ws connection?

src/rpc.rs
```diff
@@ -1,103 +1,378 @@
...
+
+	async fn current_client(&self) -> avail::Client {
+		self.client.read().await.0.clone()
+	}
+
+	pub async fn current_node(&self) -> Node {
+		self.client.read().await.1.clone()
+	}
+
+	pub async fn with_client<F, Fut, T>(&self, mut f: F) -> Result<T>
```
luka-ethernal, line 159: It would be nice if this function is rewritten for better comprehension, it has lots of moving parts, it's difficult to understand. Some comments would also not be bad. Also, this is central to the entire PR, this function needs to be in the best possible state. Let's spend some time improving this.

src/rpc.rs
i1i1: I think it is not valid if it failed during the response

src/rpc.rs
```diff
@@ -1,103 +1,378 @@
...
+			let mut f = f;
+			let mut nodes = self.nodes.clone();
+			Self::shuffle_full_nodes(&mut nodes, Some(&last_node.host));
+			let (client, node) = Self::connect_to_available_rpc(
+				&nodes,
+				&self.expected_version,
+				Some(last_node.genesis_hash),
+			)
+			.await
+			.map_err(backoff::Error::permanent)?;
```
i1i1, line 182: Maybe this error should also be transient here. WDUT?

src/rpc.rs
```diff
@@ -1,103 +1,378 @@
...
+
+			match f(client.clone()).await {
+				Ok(ok) => Ok((ok, (client, node))),
```
luka-ethernal, line 185: Are we retrying on any kind of error here? The error might not be related to connection or internal node issues? We might think of a way to differentiate among different kinds of errors, perhaps even wrapping subxt errors into our own ones.

src/rpc.rs
i1i1: Can you suggest how it can be improved? I can add comments describing what I do here, but apart from that I don't know how to improve it :man_shrugging:

src/rpc.rs
i1i1: Yes, on all errors here. Check out the error variants in subxt: https://docs.rs/subxt/latest/subxt/error/enum.Error.html#variants

I think every one of those is the reason we want to switch our RPC node

src/rpc.rs
i1i1: Added comments

src/rpc.rs
```diff
@@ -1,103 +1,387 @@
...
+			.await
+			.map_err(backoff::Error::transient)?;
+
+			match f(client.clone()).await {
+				// If closure works then we found a new working full node
+				Ok(ok) => Ok((ok, (client, node))),
+				Err(error) => {
+					warn!(%error, "Failed to connect to node. Trying to reach to another one");
+					Err(backoff::Error::transient(anyhow::Error::from(error)))
+				},
```
i1i1, line 193: Merge it together

src/rpc.rs
```diff
@@ -1,103 +1,387 @@
...
+				warn!(%error, "Failed to connect to node. Trying to reach to another one");
+			},
+		}
+		let last_node = self.current_node().await;
+
+		let (ok, client) = backoff::future::retry(self.backoff.clone(), || async {
+			let mut f = f;
+			let mut nodes = self.nodes.clone();
+			// Shuffle nodes discarding the last one
+			Self::shuffle_full_nodes(&mut nodes, Some(&last_node.host));
```
i1i1, line 177: Handle case of 1 node

src/rpc.rs
i1i1: `with_client` uses it, so resolving this one

src/rpc.rs
```diff
@@ -1,103 +1,387 @@
...
+				};
+
+				loop {
+					// We need to return, as stream has ended
+					let Some(result) = stream.next().await else { return };
+					// If we received error that means that we need to find a new client
+					let Ok(res) = result else { break };
+					yield Ok(res);
+				}
+			}
```
i1i1, line 235:
```suggestion
			loop 'outer {
				let mut stream = match self.with_client(f).await {
					Ok(s) => s,
					Err(err) => {
						yield Err(err);
						return;
					}
				};

				loop {
					// We need to return, as stream has ended
					let Some(result) = stream.next().await else { return };
					// If we received error that means that we need to find a new client
					let Ok(res) = result else { continue 'outer };
					yield Ok(res);
				}
			}
```

src/rpc.rs
```diff
@@ -1,103 +1,387 @@
...
+
+	async fn current_client(&self) -> avail::Client {
+		self.client.read().await.0.clone()
+	}
+
+	pub async fn current_node(&self) -> Node {
+		self.client.read().await.1.clone()
+	}
+
+	pub async fn with_client<F, Fut, T>(&self, mut f: F) -> Result<T>
```
i1i1, line 159: Remove `pub`

src/rpc.rs
```diff
@@ -1,103 +1,570 @@
...
+				Ok(ok) => return Ok(Either::Left(ok)),
+				Err(error) => {
+					warn!(%error, "Failed to connect to node. Trying to reach to another one");
+				},
+			}
+			let last_node = self.current_node().await;
+
+			let mut nodes = self.nodes.clone();
+			// Shuffle nodes discarding the last one
+			Self::shuffle_full_nodes(&mut nodes, Some(&last_node.host));
```
aterentic-ethernal, line 202: This doesn't discard the last one, its only pushing it to the end of a list. However, since next call is parallel, this call is not needed.

src/rpc.rs
```diff
@@ -1,103 +1,570 @@
...
+		};
+		let (client, node, ()) = Self::connect_to_available_rpc_and_return_call(
+			&nodes,
+			&expected_version,
+			expected_genesis_hash,
+			|_| futures::future::ok(()),
+		)
+		.await?;
+
+		info!(?node.genesis_hash);
```
aterentic-ethernal, line 159: Can we add some context message here?

aterentic-ethernal (commented):
I still think its not ok to spam entire network in case if some node goes down. It seems that simplest and most common case would be to connect to the next node in the list. Also last node should be used as starting point, not as priority.
This probably implies following algorithm:
  - Load `last node` at the beginning, and shuffle the list
  - Connect to the `last node` and update last node in the database, if needed
  - In case of RPC or WS failure, move the last node from `nodes` to `ignored_nodes` (or better named) list and go throught rest of the nodes, node by node
    - If connection and RPC is successful, update last node in the database and continue (if following simplifies algorithm, we can make RPC in the next loop iteration)
    - In the case of failure we go to the next node, repeating entire process
    - If list is empty we move all nodes from the `ignored_nodes` to `nodes`, shuffle the list again and repeat entire process using configured backoff
    - If we still cannot connect, lc exits
This should be fairly simple but we need to test it througly since its essential feature, so writing unit test for abstracted algorithm would be best option. Otherwise, we need to think how to test this properly, before merging this PR.

src/rpc.rs
i1i1: It was like `info!("Genesis hash: {genesis_hash}")` before, so I kept it like this. What should it be?

i1i1 (commented):
> I still think its not ok to spam entire network in case if some node goes down. It seems that simplest and most common case would be to connect to the next node in the list. Also last node should be used as starting point, not as priority. This probably implies following algorithm:

Can we keep it simple for now? I can remove the parallelization part so that network won't be flooded.

I think we agreed at the beginning that this pr is a starting point so that we won't need to bother with banning or ignoring nodes.

>  This should be fairly simple but we need to test it througly since its essential feature, so writing unit test for abstracted algorithm would be best option. Otherwise, we need to think how to test this properly, before merging this PR.

@luka-ethernal and I both tested this pr locally (Idk about Luka, but I've tested it with 2 nodes and 1 LC setup) and it worked.

sh3ll3x3c (changes requested):
We definitely need to remove the node spam and implement proper sequential (re)connect logic instead of blindly firing at all the nodes from the list.   

Aside from that, the overall approach @aterentic-ethernal  mentioned above (and the rest of the team agreed upon) is more concise and less complex than the one in this PR. To expand on it a bit, we should introduce a sequential event handler for Subxt client (a producer) that handles just the clients Websocket connection and receives commands (using channels) from all 3 clients - ie. light client, app client and sync client. Specific node-side RPC errors need to be handled accordingly.

The event handler should abstract away the reconnection algorithm, the specifics steps of it you should propose and we should then discuss it before going forward with an actual implementation.

Also - this is a huge change with potentially unforeseen consequences. We need to double down on testing and make sure this really works before we consider merging anything. With the proposed changes and some smart mocking, unit tests need to be put in place to cover at least the reconnect logic (which should be sufficiently decoupled from side-effects) as well as at least some edge cases we can think about.

i1i1:
> definitely need to remove the node spam

It is very hypothetical that it will be, but I've pushed commit which fixes it.

> the overall approach ... is more concise and less complex than the one in this PR

In short, I think this is not true.

Also, as I've mentioned before, we've talked with @aterentic-ethernal, and we agreed on this algorithm with slight changes which I've already addressed. I don't see any new properties, this new algorithm brings.

So if there are no other critical things to address, I would prefer to keep it as it is and merge it like that. 

> To expand on it a bit, we should introduce a sequential event handler for Subxt client (a producer) that handles just the clients Websocket connection and receives commands (using channels) from all 3 clients - ie. light client, app client and sync client.

It is already a pretty long pr, let's just keep it simple and address all these things separately

> Specific node-side RPC errors need to be handled accordingly.

Take a look at this comment: https://github.com/availproject/avail-light/pull/278#discussion_r1319949555

> Also - this is a huge change with potentially unforeseen consequences. We need to double down on testing and make sure this really works before we consider merging anything. With the proposed changes and some smart mocking, unit tests need to be put in place to cover at least the reconnect logic (which should be sufficiently decoupled from side-effects) as well as at least some edge cases we can think about.

We've iterated on this with Alex and there is no good way to write some test apart from rewriting everything and doubling code size with mocks which only partly useful. 

A better approach would be to have an integration test framework which would be able to spin up full nodes. This way we would have more meaningful tests.

But anyway it is out of scope of this pr. As I've said me and @luka-ethernal already tested some scenarios and it everything works.

src/rpc.rs
```diff
@@ -1,103 +1,559 @@
...
+					}
+				};
+
+				loop {
+					// We need to return, as stream has ended
+					let Some(result) = stream.next().await else { return };
+					// If we received error that means that we need to find a new client
+					let Ok(res) = result else { continue 'outer };
+					yield Ok(res);
+				}
```
i1i1, line 247: I've retested everything, and it seems that during the connection drop, instead of sending an error rpc, jsonrpsee subscription sends `None` which is weird.

So I'll revert it back to:
```rust
while let Some(Ok(res)) = stream.next().await {
    yield Ok(res);
}
```

i1i1: Or if we know that some subscriptions are endless, we can make other workaround

PR #281, opened 2023-09-07
Api v2 submit ws

aterentic-ethernal:


src/api/v2/types.rs
```diff
@@ -335,3 +356,36 @@ pub fn handle_result(result: Result<impl Reply, impl Reply>) -> impl Reply {
...
 }
+
+#[derive(Serialize, Deserialize)]
+#[serde(tag = "topic", rename_all = "kebab-case")]
+pub enum WsResponse {
+	Error(Error),
+	Version(Response<Version>),
+	Status(Response<Status>),
+	DataTransactionSubmitted(Response<SubmitResponse>),
+}
```
i1i1, line 367: I think error is not a response. Error is an error. `Result` implements `Serialize` and `Deserialize` I think, we can just return `Result<WsResponse, Error>` instead.

Or as an alternative we can create a wrapper around the result which would return [untagged](https://serde.rs/enum-representations.html#untagged) result, so that we won't need to change API.

src/api/v2/types.rs
```diff
@@ -335,3 +356,36 @@ pub fn handle_result(result: Result<impl Reply, impl Reply>) -> impl Reply {
...
+	fn from(value: Response<SubmitResponse>) -> Self {
+		WsResponse::DataTransactionSubmitted(value)
+	}
+}
+
+impl From<Error> for WsResponse {
+	fn from(value: Error) -> Self {
+		WsResponse::Error(value)
+	}
+}
```
i1i1, line 391: I think you can introduce `derive_more` crate and derive `From` and also `TryFrom` for each enum variant.

src/api/v2/types.rs
```diff
@@ -237,29 +246,32 @@ impl Reply for SubscriptionId {
...
+	Submit,
 }
 
 #[derive(Deserialize)]
 pub struct Request {
 	#[serde(rename = "type")]
 	pub request_type: RequestType,
 	pub request_id: String,
+	pub message: Option<Transaction>,
 }
```
i1i1, line 258: Why not to have request as externally tagged enum:
```rust
#[derive(Deserialize)]
#[serde(tag = "type")]
enum RequestInner {
    Version,
    Status,
    Submit {
        message: Transaction,
    },
}

#[derive(Deserialize)]
pub struct Request {
    pub request_id: String,
    #[serde(flatten)]
    pub inner: RequestInner,
}
```

src/api/v2/types.rs
```diff
@@ -237,29 +246,32 @@ impl Reply for SubscriptionId {
...
 	Version,
 	Status,
+	Submit,
 }
 
 #[derive(Deserialize)]
 pub struct Request {
 	#[serde(rename = "type")]
 	pub request_type: RequestType,
 	pub request_id: String,
```
i1i1, line 256: It is a good idea to use [`uuid::Uuiid`](https://docs.rs/uuid/1.4.1/uuid/struct.Uuid.html) here instead of string also. Generally, it is better to avoid strings as much as possible in types.

src/api/v2/ws.rs
```diff
@@ -44,31 +48,38 @@ pub async fn connect(
...
+							"Error handling web socket message: Failed to parse request: {error}"
+						)))
 					},
 				} {
-					if let Err(error) = sender.send(Ok(Message::text::<String>(error.into()))) {
-						error!("{error}");
+					if let Err(error) = send_response(sender.clone(), error.into()) {
+						error!("Failed to send error: {error}");
 					};
 				};
```
i1i1, line 84: It became a bit hard to follow this. It is better to remove nesting like this:
```rust
let message = match result {
	Err(error) =>error!("Error receiving client message: {error}"),
    Ok(message) if !message.is_text() => continue,
    Ok(message) => message,
};
let request = match Request::try_from(message) {
    Err(error) => {
    	warn!("Error handling web socket message: {error}");
		let error = Error::bad_request_unknown(&format!(
			"Error handling web socket message: Failed to parse request: {error}"
		));
		if let Err(error) = send_response(sender.clone(), error.into()).await {
			error!("Failed to send error: {error}");
		}
		continue;
    },
	Ok(request) => request,
};

let response = match handle_request(
	request,
	&version,
	&config,
	&node,
	submitter.clone(),
	state.clone(),
)
.await
{
	Ok(response) => response,
	Err(error) => {
		error!("Error handling web socket message: {error}");
		Error::internal_server_error().into()
	}
};

if let Err(error) = send_response(sender.clone(), response) {
	error!("Failed to send response: {error}");
}
```

i1i1 (commented):
Otherwise, LGTM!

src/api/v2/types.rs
aterentic-ethernal: I will check, but keep in mind that this error is just another web socket message sent to client. Maybe we can represent both server and client errors with the same abstraction and use it in control flow, but Ill have to think about it.

i1i1: Sure! What I mean is that it would be cool to just have error prone type separate from error and use `Result` as much as possible

src/api/v2/ws.rs
aterentic-ethernal: Sure, its a complext flow, its more clear now.

i1i1: Thanks!

src/api/v2/types.rs
aterentic-ethernal: It would be great to ensure that every variant contains `message` field of different type if its needed, but Ill leave that for a moment when case appears.

src/api/v2/types.rs
```diff
@@ -244,19 +244,19 @@ impl Reply for SubscriptionId {
...
 	Status,
-	Submit,
+	Submit { message: Transaction },
 }
 
 #[derive(Deserialize)]
 pub struct Request {
-	#[serde(rename = "type")]
+	#[serde(flatten)]
 	pub request_type: RequestType,
```
i1i1, line 258: I think this is no longer request type, but rather request itself. Can you rename it?

luka-ethernal (approved):

i1i1 (approved):

momosh-ethernal (approved):

sh3ll3x3c (approved):

PR #283, opened 2023-09-15
API V2: send header verified message

aterentic-ethernal:


src/api/v2/types.rs
```diff
@@ -193,44 +200,231 @@ impl Reply for Status {
...
+		serializer.serialize_str(&hex_string)
+	}
+}
+
+impl<'de> Deserialize<'de> for Commitment {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+	where
+		D: Deserializer<'de>,
+	{
+		const LEN: usize = config::COMMITMENT_SIZE * 2 + 2;
```
sh3ll3x3c, line 270: Should we maybe explain the magic number(s) here

src/bin/avail-light.rs
```diff
@@ -335,32 +344,40 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
...
-	let (message_tx, message_rx) = channel::<(Header, Instant)>(128);
-
-	tokio::task::spawn(avail_light::subscriptions::finalized_headers(
-		rpc_client.clone(),
-		message_tx,
-		error_sender.clone(),
-		state.clone(),
-		db.clone(),
-	));
+	let (message_tx, message_rx) = broadcast::channel::<(Header, Instant)>(128);
```
sh3ll3x3c, line 347: Why change from `mpsc` to `broadcast` here? "Who" else is receiving from the `message_rx`?

src/api/v2/types.rs
```diff
@@ -193,44 +200,231 @@ impl Reply for Status {
...
+		self.subscription.topics.contains(topic)
+	}
 }
 
-pub type Clients = Arc<RwLock<HashMap<String, Client>>>;
+#[derive(Clone)]
+pub struct WsClients(pub Arc<RwLock<HashMap<String, WsClient>>>);
+
+impl WsClients {
+	pub async fn set_sender(&self, subscription_id: &str, sender: Sender) -> bool {
```
sh3ll3x3c, line 389: Should this func return a `Result` maybe?

src/bin/avail-light.rs
aterentic-ethernal: One is the light client which starts with the random sampling when header is verified, and another is task that sends header verified message to connected web socket clients. I guess that confusion comes from broadcast API which creates another receiver by subscribint to the sender:

```rust
message_tx.subscribe()
```

instead of cloning the sender.

src/api/v2/types.rs
aterentic-ethernal: It kind of awkward since caller checks if subscription exists and than tries to set it. Lock is not preserved between those two calls. But to make it more clear, I'll change it to result and log error instead of info. I guess this could be done better 🤷

src/api/v2/types.rs
aterentic-ethernal: Done.

PR #284, opened 2023-09-15
sync_client goes crazy if finalized head height is 0

luka-ethernal:
Fix a cornercase where if current head height is 0, then the sync end_block will be u32::MAX


momosh-ethernal (approved):
Love the PR headline 😁

sh3ll3x3c (approved):

PR #286, opened 2023-09-18
4 new metric attributes added

sh3ll3x3c:
Added 4 new metric attributes:
- `ip` -> IP address of the light client taken from it's external multi-address 
- `job` -> set to `avail_light_client`; job is a default OT metric attribute for every metric sent
- `version` -> Light Client version retrieved from CARGO_PKG_VERSION env variable
- `role` -> type of the client: `crawler`, `app`, `bootstrap`, `relay` or `lightclient`

### DIFF
--- a/src/bin/avail-light.rs
+++ b/src/bin/avail-light.rs
@@ -160,9 +160,14 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 		while let Some(network_dump_event) = network_stats_receiver.recv().await {
 			// Set multiaddress for metric dispatch
 			if !network_dump_event.current_multiaddress.is_empty() {
-				*network_stats_metrics.multiaddress.write().unwrap() =
-					network_dump_event.current_multiaddress;
-				*network_stats_metrics.ip.write().unwrap() = network_dump_event.current_ip;
+				*network_stats_metrics
+					.multiaddress
+					.write()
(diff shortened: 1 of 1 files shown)


src/bin/avail-light.rs
```diff
@@ -113,7 +113,8 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 		tracing::subscriber::set_global_default(default_subscriber(log_level))
 			.expect("global default subscriber is set")
 	}
-
+	let version = env!("CARGO_PKG_VERSION");
```
aterentic-ethernal, line 116: I think you can use `clap::crate_version!()` for this value.

src/bin/avail-light.rs
```diff
@@ -147,6 +162,7 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 			if !network_dump_event.current_multiaddress.is_empty() {
 				*network_stats_metrics.multiaddress.write().unwrap() =
 					network_dump_event.current_multiaddress;
+				*network_stats_metrics.ip.write().unwrap() = network_dump_event.current_ip;
```
aterentic-ethernal, line 165: Maybe just to put expect here instead of unwrap?

aterentic-ethernal (approved):

src/bin/avail-light.rs
sh3ll3x3c: Fixed, tnx.

PR #287, opened 2023-09-20
API V2: send confidence achieved message

aterentic-ethernal:


momosh-ethernal (approved):

src/api/v2/mod.rs
```diff
@@ -96,43 +96,41 @@ fn ws_route(
...
 
-		let topic = Topic::HeaderVerified;
-		let message = PublishMessage::HeaderVerified(message);
-		publish_message(&clients, topic, message).await;
+		match clients.publish(&topic, message).await {
+			Ok(results) => {
+				let published = results.iter().filter(|&result| result.is_ok()).count();
+				let failed = results.iter().filter(|&result| result.is_err()).count();
+				info!(?topic, published, failed, "Message published to clients");
+				for error in results.into_iter().filter_map(Result::err) {
```
sh3ll3x3c, line 128: In case of multiple ws clients connecting simultaneously, should we think about aggregating the errors into a single `warn` message?

sh3ll3x3c (approved):

src/api/v2/mod.rs
aterentic-ethernal: Switched to `debug`.

PR #323, opened 2023-10-23
Introduce additional flags

sh3ll3x3c:
- `port`, `clean` and `verbosity` flags added
- Minor error messaging changes and README update


momosh-ethernal (approved):

src/types.rs
```diff
@@ -602,25 +612,48 @@ impl Network {
 	}
 }
 
+#[derive(Clone)]
```
aterentic-ethernal, line 615: I think you can just use serde here, with case specification.

aterentic-ethernal (approved):

PR #327, opened 2023-10-27
Reduce code duplication by abstracting away sync command execution.

aterentic-ethernal:


sh3ll3x3c (approved):
lgtm, no apparent performance degradation

PR #328, opened 2023-10-27
Pre-generated SRS loading 

ToufeeqP:
Companion PRs:
avail-core [PR#58](https://github.com/availproject/avail-core/pull/58)
avail [PR#270](https://github.com/availproject/avail/pull/270)


src/bin/avail-light.rs
```diff
@@ -192,7 +192,7 @@ async fn run(error_sender: Sender<anyhow::Error>) -> Result<()> {
 	#[cfg(feature = "network-analysis")]
 	tokio::task::spawn(network_analyzer::start_traffic_analyzer(cfg.port, 10));
 
-	let pp = Arc::new(kate_recovery::testnet::public_params(1024));
+	let pp = Arc::new(kate_recovery::testnet_v2::public_params());
```
prabal-banerjee, line 195: Let's use the Couscous?

Cargo.lock
```diff
@@ -442,13 +442,13 @@ dependencies = [
  "parity-scale-codec",
  "scale-info",
  "serde",
- "sp-arithmetic 16.0.0",
- "sp-core 21.0.0",
- "sp-io 23.0.0",
+ "sp-arithmetic 7.0.0",
```
aterentic-ethernal, line 445: I think we should revert changes which reduces versions to old ones, it seems that every update switches to old/new versions randomly.

aterentic-ethernal (approved):
Also, I think we should update public parameters in test, but not 100% sure.

Cargo.lock
ToufeeqP: Ah, I see, No idea why this is happening. but will revert it

PR #329, opened 2023-10-31
Bootstrap peer dial check

sh3ll3x3c:
- Added a peer dial feature for a specific peerID, that checks if the peer is reachable or not
- Bootstrap process now fails if the bootstrap peer(s) in the list aren't reachable

### DIFF
--- a/src/network/p2p/client.rs
+++ b/src/network/p2p/client.rs
@@ -8,6 +8,7 @@ use kate_recovery::{
 use libp2p::{
 	kad::{record::Key, PeerRecord, Quorum, Record},
 	multiaddr::Protocol,
+	swarm::DialError,
 	Multiaddr, PeerId,
 };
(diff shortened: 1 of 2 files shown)


aterentic-ethernal (approved):

src/network/p2p/client.rs
```diff
@@ -376,6 +400,11 @@ pub enum Command {
 		peer_addr: Multiaddr,
 		response_sender: oneshot::Sender<Result<()>>,
 	},
+	DialAddress {
+		peer_id: PeerId,
+		peer_addr: Multiaddr,
+		response_sender: oneshot::Sender<Result<(), DialError>>,
```
momosh-ethernal, line 406: use anyhow::Error here?

momosh-ethernal (approved):
All good, check comment for error type

PR #331, opened 2023-11-01
change types regards to the header change

kroos47:
Changing the types regards to the header change


sh3ll3x3c (approved):
lgtm, but lets wait for the node release before we merge

PR #345, opened 2023-11-15
Luka/gen hash fix

luka-ethernal:
- Fix genesis_hash storing and checking (also system_version and spec_version)
- Add more useful logs in case finality verification fails

### DIFF
--- a/src/network/rpc/event_loop.rs
+++ b/src/network/rpc/event_loop.rs
@@ -301,9 +301,14 @@ impl EventLoop {
 							&signed_message,
 							&precommit.id,
 						);
-						is_ok
-							.then(|| precommit.clone().id)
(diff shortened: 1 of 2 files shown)


src/network/rpc/event_loop.rs
```diff
@@ -138,6 +139,9 @@ impl EventLoop {
 			"Connection established to the Node: {:?} <{version}>",
 			node.host
 		);
+		self.nodes.get_current_mut().spec_version = runtime_version.spec_version;
```
aterentic-ethernal, line 142: `let current node = self.nodes.get_current_mut();` and then rest of the code?

aterentic-ethernal (approved):

sh3ll3x3c (approved):

PR #358, opened 2023-11-22
Remove disable proof verification configuration

aterentic-ethernal:
Since `disable_proof_verification` description is misleading and only
use of this parameter is to disable proof verification during the
confidence check, we can remove this parameter.

### DIFF
--- a/src/light_client.rs
+++ b/src/light_client.rs
@@ -223,34 +223,31 @@ pub async fn process_block(
 		return Ok(None);
 	}
 
-	let mut confidence = None;
-	if !cfg.disable_proof_verification {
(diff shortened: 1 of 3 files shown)


luka-ethernal (approved):

kroos47 (approved):

prabal-banerjee:
But what do we use instead to fetch cells without verifying proof?

aterentic-ethernal:
> But what do we use instead to fetch cells without verifying proof?

This covers only random sampling cells, which is insignificant.

momosh-ethernal (approved):

PR #361, opened 2023-11-22
Fetch unverified cells from rpc

aterentic-ethernal:


src/network/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+		let (verified, mut unverified) = proof::verify(
+			block_number,
+			dimensions,
+			&dht_fetched,
+			commitments,
+			self.pp.clone(),
+		)
+		.context("Failed to verify fetched cells")?;
+
+		info!(
```
sh3ll3x3c, line 91: We should add a separate metric of the time it took to fetch & verify cells from DHT.

sh3ll3x3c (approved):

kroos47 (approved):

PR #369, opened 2023-11-27
Decouple async put

sh3ll3x3c:
- Made DHT cell upload totally async
- Removed bathing from DHT puts
- Moved success rate calculation to the event handling part of the event loop
- PUT success rate for previous block is calculated when new block resets the cell counter to 0


src/light_client.rs
```diff
@@ -283,10 +285,21 @@ pub async fn process_block(
 					.collect::<Vec<_>>();
 				rpc_fetched.extend(partition_fetched_filtered.clone());
 			}
```
sh3ll3x3c, line 287: Moved RPC timing calculation here.

src/light_client.rs
```diff
@@ -283,84 +273,45 @@ pub async fn process_block(
...
+			let rpc_call_duration = begin.elapsed();
```
aterentic-ethernal, line 276: Wouldn't this always measure aggregated time from the begining of the loop?

src/light_client.rs
```diff
@@ -283,84 +273,45 @@ pub async fn process_block(
...
-
-		metrics
-			.record(MetricValue::DHTPutRowsDuration(time_elapsed.as_secs_f64()))
-			.await?;
+			.await
+			.map_err(|e| {
+				warn!("Error inserting rows into DHT: {e}");
+				e
+			})
+			.ok();
```
aterentic-ethernal, line 304: Since we are not returning error, we can use `if let Err(error) = ...` here, to avoid confusion. But previous code was returning error, so is this intentional?

src/light_client.rs
sh3ll3x3c: Yeah, the idea is to measure total time needed to retrieve all of the data from the node.

src/light_client.rs
sh3ll3x3c: Yeah, we wan't to log the errors but we don't want to stop the block processing loop because of PUT errors. At least for now...

momosh-ethernal (approved):

src/network/p2p/client.rs
```diff
@@ -524,37 +510,25 @@ impl Client {
...
-				Ok(DHTPutSuccess::Batch(num)) => num_success += num,
-				Err(err) => return Err(err),
-			};
-		}
-		Ok(DHTPutSuccess::Batch(num_success))
+		block_num: u32,
+	) -> Result<()> {
+		let block_size = records.len();
+
+		// create oneshot for each chunk, through which will success counts be sent,
```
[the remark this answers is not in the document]
sh3ll3x3c: yeah, I sometimes love to warm them up in the microwave 🍽️

PR #382, opened 2023-12-12
added external peer filtering on swarm and kademlia levels

sh3ll3x3c:
- Swarm level filtering added using `Identify` protocol, with a temporary peer blocking for external non-Avail peers
- Re-introduced the use of Kademlia protocol names as a secondary protection, preventing future merging of Avail Kademlia with outside networks

### DIFF
--- a/src/network/p2p/event_loop.rs
+++ b/src/network/p2p/event_loop.rs
@@ -69,6 +69,12 @@ struct BootstrapState {
 	timer: Interval,
 }
 
+// Identity strings used for peer filtering
+pub struct IdentityData {
+	pub agent_version: String,
+	pub protocol_version: String,
+}
+
 pub struct EventLoop {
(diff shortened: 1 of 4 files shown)


src/network/p2p/event_loop.rs
```diff
@@ -276,9 +284,22 @@ impl EventLoop {
...
+						info:
+							Info {
+								listen_addrs,
+								agent_version,
+								protocol_version,
+								..
+							},
 					} => {
 						trace!("Identity Received from: {peer_id:?} on listen address: {listen_addrs:?}");
+						if agent_version != self.identity_data.agent_version
```
momosh-ethernal, line 296: consider adding an else branch, if all matched to add a peer to kad

aterentic-ethernal (approved):

momosh-ethernal (approved):

PR #397, opened 2023-12-22
Added Kademlia mode check before inserting peer to routing table

sh3ll3x3c:
- Identify protocols  `agent_version` is used for propagating LCs Kademlia mode
- On identify received event, only peers in `server` mode are inserted to the routing table
- Switched the default Kademlia mode to `client`
- `agent_version` is now in the `{base_string}/{client_type}/{kademlia_mode}` format

### DIFF
--- a/src/network/p2p/event_loop.rs
+++ b/src/network/p2p/event_loop.rs
@@ -16,7 +16,7 @@ use libp2p::{
 	upnp, Multiaddr, PeerId, Swarm,
 };
 use rand::seq::SliceRandom;
-use std::{collections::HashMap, sync::Arc, time::Duration};
+use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
 use tokio::{
 	sync::oneshot,
 	time::{interval_at, Instant, Interval},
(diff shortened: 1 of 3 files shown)


src/network/p2p/event_loop.rs
```diff
@@ -304,47 +304,58 @@ impl EventLoop {
...
+							listen_addrs,
+							agent_version,
+							protocol_version,
+							..
+						},
+				} => {
+					trace!(
+						"Identity Received from: {peer_id:?} on listen address: {listen_addrs:?}"
+					);
+					let incoming_peer_agent_version = match AgentVersion::from_str(&agent_version) {
```
momosh-ethernal, line 323: use let else here?

src/network/p2p/event_loop.rs
```diff
@@ -304,47 +304,58 @@ impl EventLoop {
...
+						Ok(agent) => agent,
+						Err(e) => {
+							info!("Error parsing incoming agent version: {e}");
+							return;
+						},
+					};
+					if protocol_version != self.identity_data.protocol_version {
```
momosh-ethernal, line 330: reverse condition? math if without negation

momosh-ethernal (approved):

src/network/p2p/event_loop.rs
sh3ll3x3c: We need both values in this case, so a match is probably a better choice...

src/network/p2p/event_loop.rs
sh3ll3x3c: Done.

luka-ethernal (approved):

PR #398, opened 2023-12-25
Fix http server spawning.

0xfluxx:


sh3ll3x3c (approved):
too complex for my understanding. lgtm

PR #407, opened 2024-01-09
fixed wrong put success rate metric calculation

sh3ll3x3c:


sh3ll3x3c:
> Can `block.total_count` achieve a value equal to zero? In that case we might need to define a success_rate value.

No, no, total count is >0 (for now) because the matrix can't be of 0 size.

PR #429, opened 2024-01-26
feat: Builds Docker image from LC bin releases

moraesjeremias:
## Changes
- [x] Builds Docker image from LC bin releases
- [x] Rename linux aarch64 bin to arm64

## Reason for Change
- Docker image building pipeline is currently broken due to missing headers during build time. Under internal discussion we come up with an idea to just download the recent published binary, bypassing header bugfix and turning the docker build process lightweight and faster (14 min compared to 20s).
- Upon this new docker image build strategy we came across some different name convention on 64 bits. `aarch64`used to be supported by Docker but [was deprecated](https://hub.docker.com/r/aarch64/ubuntu) in favor of `arm64v8`. So in order to match the [platform supported by Docker](https://github.com/docker-library/official-images#architectures-other-than-amd64) this just change the binary name, but continue using the `aarch64` GNU tooling for building it.


.github/workflows/releaser.yml
```diff
@@ -43,12 +43,12 @@ jobs:
...
             mv target/aarch64-unknown-linux-gnu/maxperf/avail-light target/aarch64-unknown-linux-gnu/maxperf/avail-light-linux-aarch64
             pushd target/aarch64-unknown-linux-gnu/maxperf/
-            tar czf avail-light-linux-aarch64.tar.gz avail-light-linux-aarch64
+            tar czf avail-light-linux-arm64.tar.gz avail-light-linux-aarch64
             popd
          - uses: actions/upload-artifact@v2
            with:
-             name: avail-light-linux-aarch64-binary
-             path: target/aarch64-unknown-linux-gnu/maxperf/avail-light-linux-aarch64.tar.gz
+             name: avail-light-linux-arm64-binary
```
sh3ll3x3c, line 50: Any chance we might add a Windows build? Only additional effort would be that it would require clang for compile...

sh3ll3x3c (approved):

.github/workflows/releaser.yml
moraesjeremias: I'd rather work on this in a separate PR.
Don't wanna couple many changes. I'll put it on the back log for upcoming week.

obrienalaribe (approved):

PR #433, opened 2024-01-31
Defragment restart

aterentic-ethernal:


momosh-ethernal (approved):

aterentic-ethernal:
Restarts doesn't reduce memory usage.

PR #439, opened 2024-02-08
479 - RPC Calls with Retries

momosh-ethernal:
The new RPC Client supports multiple configurations for Retry strategies. If transient errors cannot be retried successfully, the RPC Client will switch to a new node and attempt the same call again. This process will be repeated until all available nodes are exhausted.

### DIFF
--- a/src/bin/api_compat_test.rs
+++ b/src/bin/api_compat_test.rs
@@ -29,8 +29,9 @@ async fn main() -> Result<()> {
 		retries: 4,
 	});
 
-	let (rpc_client, _, event_loop) = rpc::init(db, state, &[command_args.url], "DEV", retry_cfg);
-	tokio::spawn(event_loop.run());
+	let (rpc_client, _, subscriptions) =
+		rpc::init(db, state, &[command_args.url], "DEV", retry_cfg).await?;
+	tokio::spawn(subscriptions.run());
 
 	let mut correct: bool = true;
 
--- a/src/network/rpc/client.rs
+++ b/src/network/rpc/client.rs
@@ -139,7 +139,7 @@ impl Client {
 		// after a successful connection, try to execute passed call predicate
 		for Node { host, .. } in nodes.iter() {
 			let result =
-				Self::create_subxt_client(&host, expected_node.clone(), expected_genesis_hash)
+				Self::create_subxt_client(host, expected_node.clone(), expected_genesis_hash)
 					.and_then(move |(client, node)| {
 						predicate(client.clone()).map_ok(|res| (client, node, res))
 					})
@@ -194,7 +194,7 @@ impl Client {
 		// create Header subscription
 		let header_subscription = client.rpc().subscribe_finalized_block_headers().await?;
 		// map Header subscription to the same type for later matching
-		let headers = header_subscription.map_ok(|h| Subscription::Header(h));
+		let headers = header_subscription.map_ok(Subscription::Header);
 
 		let justification_subscription = client
 			.rpc()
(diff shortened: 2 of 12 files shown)


src/bin/api_compat_test.rs
```diff
@@ -29,8 +29,9 @@ async fn main() -> Result<()> {
 		retries: 4,
 	});
 
-	let (rpc_client, _, event_loop) = rpc::init(db, state, &[command_args.url], "DEV", retry_cfg);
-	tokio::spawn(event_loop.run());
+	let (rpc_client, _, subscriptions) =
+		rpc::init(db, state, &[command_args.url], "DEV", retry_cfg).await?;
+	tokio::spawn(subscriptions.run());
```
aterentic-ethernal, line 34: I guess we should shutdown light client here if run fails?

src/network/rpc/client.rs
```diff
@@ -1,612 +1,289 @@
...
+		// check genesis hash
+		let genesis_hash = client.genesis_hash();
+		info!("Genesis hash: {:?}", genesis_hash);
+		if let Some(cfg_genhash) = from_hex(expected_genesis_hash)
+			.ok()
+			.and_then(|e| TryInto::<[u8; 32]>::try_into(e).ok().map(H256::from))
+		{
+			if !genesis_hash.eq(&cfg_genhash) {
+				Err(eyre!(
+					"Genesis hash doesn't match the configured one! Change the config or the node url ({}).", host
```
sh3ll3x3c, line 90: We should add to log that we're skipping to the next node.

src/network/rpc/client.rs
```diff
@@ -1,612 +1,289 @@
...
+				))?
+			}
+		} else if expected_genesis_hash.starts_with(DEV_FLAG_GENHASH) {
+			warn!("Genesis hash configured for development ({}), skipping the genesis hash check entirely.", expected_genesis_hash);
+		} else {
+			Err(eyre!(
+				"Genesis hash invalid, badly configured or missing (\"{}\").",
```
sh3ll3x3c, line 97: Same - we should add to log that we're skipping over to the next node (or exiting client).

src/network/rpc/client.rs
```diff
@@ -1,612 +1,289 @@
...
-		// TODO: consider what to do if this results with None
-		// TODO: maybe wrap error with specific error message per Command type implementation
-		if let Some(sender) = self.response_sender.take() {
-			_ = sender.send(Err(error));
+	async fn with_retries<F, Fut, T>(&self, mut predicate: F) -> Result<T>
+	where
+		F: FnMut(avail::Client) -> Fut + Copy,
+		Fut: std::future::Future<Output = Result<T, subxt::error::Error>>,
+	{
+		// try and execute the passed function, use the Retry strategy if needed
```
sh3ll3x3c, line 162: We should try to log every iteration of retries and timeouts, so that the UX is more responsive towards the user. 

e.g. Retrying X 1/8... Waiting 3s... etc

src/network/rpc/client.rs
```diff
@@ -690,43 +390,62 @@ impl Client {
...
+		signer: &PairSigner<AvailConfig, Pair>,
+		other_params: avail_subxt::primitives::AvailExtrinsicParams,
+	) -> Result<subxt::blocks::ExtrinsicEvents<AvailConfig>> {
+		let tx_progress = self
+			.with_retries(|client| {
+				let other_params = other_params.clone();
+				async move {
+					client
+						.tx()
+						.sign_and_submit_then_watch(call, signer, other_params)
```
sh3ll3x3c, line 416: Did we consider that there might be multiple undesired submtis if we do retries here?

sh3ll3x3c (approved):

PR #443, opened 2024-02-12
Luka/support multi versions

luka-ethernal:


momosh-ethernal (approved):

sh3ll3x3c (approved):

sh3ll3x3c:
Lets make this the first step, with the idea of moving supported node version to config.

PR #452, opened 2024-02-18
feat: add ping latency measurement.

vbhattaccmu:
The following PR adds ping latency to the existing set of metrics on light client telemetry. This will enable us to accurately monitor and correlate round trip times for messages exchanged, with existing network traffic conditions.


sh3ll3x3c (approved):
lgtm!

PR #470, opened 2024-03-05
Optional header extension support.

aterentic-ethernal:


aterentic-ethernal:
Closed since its contained in https://github.com/availproject/avail-light/pull/536.

PR #480, opened 2024-03-29
fix: change binary name inside linux arm64 tarball

QEDK:
- [x] Fixes binary name to match tarball name


moraesjeremias (commented):
LGTM

moraesjeremias (approved):

PR #528, opened 2024-04-09
Fix mode in case when app id is set to zero in configuration.

aterentic-ethernal:
This PR fixes issue with configuration which contains `app_id = 0` setting. This configuration runs application client with app_id 0 which is not desired behavior. Fix uses `Mode` to decide if application client should be run. 

Alternative that should be considered should parse configuration into enum which is a better representation of the setting:

```rust
pub enum AppId {
	None,
	System,
	Application(avail_core::AppId),
}
```

But since changes are not trivial, this will be implemented in the future PRs.

### DIFF
--- a/src/bin/avail-light.rs
+++ b/src/bin/avail-light.rs
@@ -11,7 +11,7 @@ use avail_light::{
 	sync_client::SyncClient,
 	sync_finality::SyncFinality,
 	telemetry::{self, otlp::MetricAttributes},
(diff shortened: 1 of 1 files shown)


aterentic-ethernal:
It seems that having application client with ID=0 is a valid case, and its reserved for avail, so Ill close this PR and create a task to improve configuration semantics, and fix will be done in availup.

PR #534, opened 2024-04-10
Use AppId type for configuration.

aterentic-ethernal:


aterentic-ethernal:
Closed as irrelevant due to spec changes.

PR #537, opened 2024-04-16
Remove multiaddress from metrics

aterentic-ethernal:
Main aim of this PR is to remove multiaddress from the metrics attributes, and introduce changes from goldberg:
- Add avail public key in hex format to logs
- ~~Send IP as metrics attribute~~
- Invoke health check metric before processing first block

### DIFF
--- a/src/maintenance.rs
+++ b/src/maintenance.rs
@@ -42,16 +42,6 @@ pub async fn process_block(
 		.await
 		.wrap_err("Unable to get Kademlia map size")?;
 
-	// Get last confirmed external multiaddress
-	if let Ok(multiaddrs) = p2p_client.get_multiaddress_and_ip().await {
-		debug!("Confirmed external multiaddresses: {:?}", multiaddrs);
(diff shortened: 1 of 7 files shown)


src/maintenance.rs
```diff
@@ -42,14 +42,8 @@ pub async fn process_block(
...
 
-	// Get last confirmed external multiaddress
-	if let Ok(multiaddrs) = p2p_client.get_multiaddress_and_ip().await {
-		debug!("Confirmed external multiaddresses: {:?}", multiaddrs);
-		if let Some(last_confirmed_ma) = multiaddrs.last() {
-			metrics
-				.set_multiaddress(last_confirmed_ma.to_string())
-				.await;
-		}
+	if let Ok(ip) = p2p_client.get_ip().await {
```
sh3ll3x3c, line 45: I would remove the IP address too, as it is also way too mutable to be in the metric attribs. But the multiaddr should def be in logs.

sh3ll3x3c (approved):

PR #542, opened 2024-04-19
Implement RocksDB kademlia store under the feature toggle

aterentic-ethernal:
This PR adds RocksDB as persistent Kademlia record store under the `kademlia-rocksdb` feature flag.

### DIFF
--- a/src/network/p2p/kad_rocksdb_store.rs
+++ b/src/network/p2p/kad_rocksdb_store.rs
@@ -11,8 +11,7 @@ use std::collections::hash_set;
 use std::iter;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use tracing::error;
-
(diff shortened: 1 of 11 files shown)


src/network/p2p/kad_rocksdb_store.rs
```diff
@@ -0,0 +1,238 @@
...
+			Ok(_) => info!("Removed {write_batch_len} records from the RocksDB store"),
+		}
+	}
+
+	// Optimizations are not implemented currently
+	pub fn shrink_hashmap(&mut self) {}
+}
+
+impl RocksDBStore {
+	pub fn get_cf(&self) -> Option<Arc<BoundColumnFamily>> {
```
sh3ll3x3c, line 131: We should prob instrument this function too, as it could introduce some performance penalties on frequent calls.

Issue #552, opened 2024-04-25
Docker container with Light Client fails to run on Mac (Apple M1 Pro) 


vlad2095:
About my system: 
<img width="241" alt="image" src="https://github.com/availproject/avail-light/assets/29329498/604f9f6e-1981-4d5d-9bc2-7b38949ec433">

Issue:
<img width="585" alt="image" src="https://github.com/availproject/avail-light/assets/29329498/c1a48744-deda-49ef-91a7-daddfbf0cbfe">

nooga:
Hey, small update on this, looks like the arm64 `availj/avail-light:v1.8.1` image contains amd64 ubuntu with an arm64 `avail-light` binary and thus the OS cannot load it. 

Running on AppleSillicon:
```
podman run --rm -it --entrypoint=/bin/bash --user=root availj/avail-light:v1.8.1
```

Let's check the binary:
```
avail@3b9823302387:/var/lib/avail-light$ ls -la /usr/local/bin
total 43740
drwxr-xr-x. 1 root  root        25 Apr 26 06:50 .
drwxr-xr-x. 1 root  root        30 Apr 16 02:03 ..
-rwxr-xr-x. 1 avail avail 44787792 Apr 26 06:39 avail-light
avail@3b9823302387:/var/lib/avail-light$ head -n 1 /usr/local/bin/avail-light
ELF���@P^�@8
```

It's there and it is an ELF file. Let's check the size:
```
avail@3b9823302387:/var/lib/avail-light$ stat /usr/local/bin/avail-light
  File: /usr/local/bin/avail-light
  Size: 44787792  	Blocks: 87480      IO Block: 4096   regular file
Device: 51h/81d	Inode: 8709281     Links: 1
Access: (0755/-rwxr-xr-x)  Uid: ( 1000/   avail)   Gid: ( 1000/   avail)
Access: 2024-04-26 06:39:55.000000000 +0000
Modify: 2024-04-26 06:39:55.000000000 +0000
Change: 2024-05-09 12:17:38.803787688 +0000
 Birth: -
```

But it doesn't run:
```
avail@3b9823302387:/var/lib/avail-light$ file /usr/local/bin/avail-light
bash: file: command not found
```

Let's check it with `file`:
```
root@bf28268203f2:/var/lib/avail-light# file /usr/local/bin/avail-light
[...]

PR #558, opened 2024-05-09
Disable building broken arm64 container image in release action

nooga:
This is a makeshift fix for #552.

As explained in the issue, the arm64 seems to be unusable as it attempts to run arm64 `avail-light` on amd64 ubuntu. 

I've had a look at https://github.com/availproject/avail/blob/main/.github/workflows/releaser_docker.yml#L43 and it seems that avail does not distribute arm64 images. This is fine since amd64 images run fine on macs. 

Please see #552 for more details.

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -174,7 +174,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: ./Dockerfile.release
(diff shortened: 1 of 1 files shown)


moraesjeremias (approved):
TYSM for the troubleshoot and contribution @nooga!

LGTM.

obrienalaribe (approved):

PR #575, opened 2024-06-12
p2p minimum protocol version check, network name

vbhattaccmu:
The following PR performs the following:- 

1. Add network name of format network:genesis_hash to metrics 
2. Check minimum_protocol_version at identify step

### DIFF
--- a/src/network/p2p/event_loop.rs
+++ b/src/network/p2p/event_loop.rs
@@ -372,7 +372,6 @@ impl EventLoop {
 							incoming_peer_agent_version.release_version
 						);
 						self.swarm.behaviour_mut().kademlia.remove_peer(&peer_id);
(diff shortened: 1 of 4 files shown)


src/network/p2p/event_loop.rs
```diff
@@ -357,13 +357,24 @@ impl EventLoop {
...
 					);
+
 					let incoming_peer_agent_version = match AgentVersion::from_str(&agent_version) {
 						Ok(agent) => agent,
 						Err(e) => {
 							debug!("Error parsing incoming agent version: {e}");
 							return;
 						},
 					};
+
```
sh3ll3x3c, line 368: We need to add this check to the MDNS handler too, in the next PR.

PR #584, opened 2024-06-18
feat: MDNS Behaviour Checks

vbhattaccmu:
The following PR performs the following:- 
1. Prevents client mode LCs from getting added to routing table at mDNS step.

### DIFF
--- a/src/network/p2p/event_loop.rs
+++ b/src/network/p2p/event_loop.rs
@@ -415,7 +415,7 @@ impl EventLoop {
 							})
 							.collect();
 
-					let _ = self.swarm.behaviour_mut().identify.push(peer_ids);
+					self.swarm.behaviour_mut().identify.push(peer_ids);
 				},
 				mdns::Event::Expired(addrs_list) => {
 					addrs_list.into_iter().for_each(|(peer_id, multiaddr)| {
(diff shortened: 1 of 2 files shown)


src/network/p2p/event_loop.rs
```diff
@@ -405,34 +405,41 @@ impl EventLoop {
...
+							.into_iter()
+							.filter(|(peer_id, multiaddr)| {
+								multiaddr
+									.to_string()
+									.contains(Protocol::P2p(*peer_id).tag())
+							})
+							.for_each(|(peer_id, multiaddr)| {
+								trace!("MDNS got peer with ID: {peer_id:#?} and Address: {multiaddr:#?}");
+
+								let _ = self.swarm.dial(multiaddr);
```
sh3ll3x3c, line 421: I'd suggest that we don't use a dial here, as it would create a ton of traffic between peers on the initial boot.

What we could do is:

1. Abstract the Identify logic that deals with the version/mode checks
2. Use the same logic here as well as on the Identify 
3. Make that new check function testable

sh3ll3x3c: This won't be doable because of the input that MDNS handler receives.

Another approach is to push the Identify from the handler to force it to trigger.

sh3ll3x3c (approved):

PR #593, opened 2024-06-21
Add tests for votes_ancestries and improve check_finality a tiny bit.

luka-ethernal:


src/finality.rs
```diff
@@ -85,11 +87,15 @@ pub fn check_finality(
 				justification
 			)
 				})
-		})
-		.collect::<Result<Vec<_>>>();
+				.into()
+		});
+
+	if !failed_verifications.is_empty() {
+		warn!("Failed verifications found: {failed_verifications:?}")
```
sh3ll3x3c, line 94: We should prob add for which block is this warn related to.

aterentic-ethernal (approved):

PR #597, opened 2024-06-25
An endpoint for public multiaddress fetching for peerIDs

vbhattaccmu:
The following PR adds an endpoint for fetching public multiaddresses for a peer.

### DIFF
--- a/src/network/p2p/client.rs
+++ b/src/network/p2p/client.rs
@@ -355,7 +355,10 @@ impl Command for GetExternalPeerInfo {
 			Ok(None)
 		} else {
 			Ok(Some(PeerInfo {
-				peer_id: self.peer_id.map(|peer| peer.to_string()).expect("Could not retrieve peerID"),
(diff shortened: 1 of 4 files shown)


src/network/p2p/client.rs
```diff
@@ -296,22 +296,54 @@ impl Command for CountKademliaPeers {
 }
 
 struct GetLocalInfo {
+	peer_id: Option<PeerId>,
 	response_sender: Option<oneshot::Sender<Result<LocalInfo>>>,
 }
 
 impl Command for GetLocalInfo {
 	fn run(&mut self, entries: EventLoopEntries) -> Result<(), Report> {
+		let (peer_id, peer_multiaddr) = match self.peer_id {
```
sh3ll3x3c, line 305: We should probably add a new command for this, in order to split the domains of peer vs external peer info.

sh3ll3x3c (approved):

PR #599, opened 2024-06-25
Kademlia mode switch

sh3ll3x3c:


src/maintenance.rs
```diff
@@ -58,22 +52,37 @@ pub async fn process_block(
...
+		// Check external reachability and local hardware availability
+		let local_info = p2p_client.get_local_info().await?;
+		if !local_info.external_listeners.is_empty()
+			&& check_system_resources(
+				maintenance_config.total_memory_gb_threshold,
+				maintenance_config.num_cpus_threshold,
+			) {
+			info!("Switching Kademlia mode to server!");
+			_ = p2p_client.change_kademlia_mode(Mode::Server);
+			*is_server = true;
```
momosh-ethernal, line 66: should we move this into db? maybe we would need this info later

momosh-ethernal: could be moved with new PR, if needed

momosh-ethernal (approved):

src/maintenance.rs
sh3ll3x3c: Client starts off in `client` mode either way, until proven it's publicly reachable and satisfies the conditions. If the Kad mode state is needed, we can most likely retrieve it using a Kademlia call. 
If this proves not to be the case - I'll move the server mode to database.

PR #664, opened 2024-08-09
Crawler refactorings

aterentic-ethernal:
Verious refactorings to support crawler separation


core/Cargo.toml
```diff
@@ -60,9 +60,12 @@ void = { workspace = true }
 warp = { workspace = true }
 
 # OpenTelemetry
+clap.workspace = true
```
momosh-ethernal, line 63: check these notations

momosh-ethernal (approved):

PR #669, opened 2024-08-13
Switching to closure-based solution for P2P Commands

momosh-ethernal:
Switching to the closure-based solution enabled us to abandon the `EventLoopEntries` structure that was referenced and copied in multiple places. While removing the coupling to the structure that is a copy of the current fields on the `EventLoop` itself, this approach provides us with other benefits:

- Simplicity: It requires minimal boilerplate code and is easy to understand and implement.
- Flexibility: Each command can easily capture exactly what it needs in its closure, without needing to define new structs for each command type.
- Type Safety: It maintains strong type safety while avoiding the complexities of trait objects with generic methods.
- Ease of Use: Creating new commands is as simple as writing a closure, which can be very intuitive and quick to implement.
- Composability: Closures can be easily composed and combined, allowing for complex behaviors to be built from simpler parts.

### DIFF
--- a/core/src/api/v2/handlers/p2p.rs
+++ b/core/src/api/v2/handlers/p2p.rs
@@ -76,7 +76,7 @@ pub struct ExternalPeerMultiaddress {
 
 pub async fn get_peer_info(p2p_client: p2p::Client) -> Result<PeerInfoResponse, Error> {
 	let local_info = p2p_client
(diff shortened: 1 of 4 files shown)


aterentic-ethernal (approved):
Much better, lets just consider removing `handle_command`.

PR #694, opened 2024-09-09
Fix broken config load from file

cortze:
# Description
I've been having some trouble setting up an `avail-light-client` to a private network, as the Node version wouldn't match the remote one when using the default `full_node_ws` RPC node.

I've been assisted by @vbhattaccmu pointing to a successful `commit` and valid RPC node endpoint. 
However, the values from the config file weren't applying to the runtime config. 

Furthermore, I also spotted that the config file should be formatted as a `TOML` file, not as a `YAML` one as the example suggests.

This PR addresses both. I'm not super familiar with Rust, so feel free to suggest any changes that would make a "better" code ✌🏽


cortze:
Any inputs on this @momosh-ethernal ?
Seems that the `ot_collector_endpoint` `config.toml` value is also hardcoded `ot_collector_endpoint: "http://otel.lightclient.hex.avail.so:4317"`

sh3ll3x3c:
Hey @cortze, thanks for the contribution and sorry for letting the PR go stale.

The current `network` flag implementation is meant for having a really quick way of connecting the light client to the network, without the user having to use config files. The hierarchy is such that the network flag overrides the config file which overrides the defaults.

For advanced uses, such as different node RPC endpoints, otel endpoints, etc. the intended use is to go through the config file without using the `network` param at all.

cortze:
I see. It looks quite counter-intuitive, as some config parameters in the `toml` file aren't configurable through CLI flags (i.e., the `ot_collector_endpoint`) which defaults to a remote `op` endpoint. Would this require setting the entire `toml` 
 variables?

sh3ll3x3c:
Not at all, just the relevant fields. All other fields not set by the config file or overridden by the CLI flags will be set to their respective default values.

I guess the most important ones would be:

```tomll
bootstraps = ['/dns/bootnode.1.lightclient.mainnet.avail.so/tcp/37000/p2p/12D3KooW9x9qnoXhkHAjdNFu92kMvBRSiFBMAoC5NnifgzXjsuiM']
full_node_ws = ["wss://avail-mainnet.public.blastapi.io/"]
genesis_hash = "b91746b45e0346cc2f815a520b9c6cb4d5c0902af848db0a80f85932d2e8276a"
```


The complete config reference can be found in the readme.

sh3ll3x3c:
We'll definitely address the configuration shortcomings in the near future, again - thanks for bringing this up!

PR #702, opened 2024-09-17
Releaser updates for e2e tests

obrienalaribe:


.github/workflows/avail-light-client.yml
```diff
@@ -1,7 +1,7 @@
 name: Avail Light Client Release
 on:
   push:
-    branches-ignore:
```
aterentic-ethernal, line 4: wouldnt this create dozens of unwanted releases?

PR #704, opened 2024-09-25
update docker release image

vbhattaccmu:


Dockerfile.release
```diff
@@ -5,17 +5,26 @@ ARG BUILDPLATFORM
 
 WORKDIR /var/lib/avail-light
 
-ARG AVAIL_TAG=v1.7.8
-ARG AVAIL_LC_BIN=https://github.com/availproject/avail-light/releases/download/$AVAIL_TAG/avail-light-linux-$TARGETARCH.tar.gz
+ARG AVAIL_TAG=v1.12.0-rc2
+
+RUN apt-get update && apt-get install -y --no-install-recommends gettext wget ca-certificates \
```
moraesjeremias, line 10: This introduces another layer in the container image, bloating the image size in the end.

Ideally we should use less `RUN` steps as possible

moraesjeremias (approved):
LGTM!

Let's just put on the docs or avail-light [README](https://github.com/availproject/avail-light/blob/3c39becfee4cea6a3c70fe8327cf2a589ee123f9/client/README.md?plain=1#L61) that the `--network` flag is required to run the container.

Congrats @vbhattaccmu !

PR #724, opened 2024-10-20
Explicitly merging  RPC Subscription Streams

momosh-ethernal:
Solving the issues where the previously merged RPC Subscription stream was exhibiting bad behavior. It continued to produce items, even though one of the streams returned None. This PR introduces:

- More explicit control over stream merge
- Easier debugging
- Ability to additionally extend functionality

Streams are polled randomly, to avoid exclusively polling one stream as always the first one.

### DIFF
--- a/core/src/network/rpc/client.rs
+++ b/core/src/network/rpc/client.rs
@@ -15,11 +15,17 @@ use color_eyre::{
 	Report, Result,
 };
 use futures::{Stream, TryStreamExt};
-use std::sync::Arc;
+use rand::Rng;
+use std::{
+	iter::Iterator,
+	pin::Pin,
+	sync::Arc,
+	task::{Context, Poll},
+};
 use tokio::sync::RwLock;
 use tokio_retry::Retry;
 use tokio_stream::StreamExt;
-use tracing::{info, warn};
+use tracing::{debug, error, info, trace, warn};
 
 use super::{configuration::RetryConfig, Node, Nodes, Subscription, WrappedProof};
 use crate::{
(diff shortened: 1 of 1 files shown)


core/src/network/rpc/client.rs
```diff
@@ -105,6 +111,87 @@ impl GenesisHash {
...
+	type Item = Result<Subscription, subxt::error::Error>;
+
+	fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+		// Randomly decide which stream to poll first
+		let poll_headers_first = rand::thread_rng().gen_bool(0.5);
+		if poll_headers_first {
+			self.poll_both(cx, true)
+		} else {
+			self.poll_both(cx, false)
+		}
```
aterentic-ethernal, line 130: Just `self.poll_both(cx, poll_headers_first)`?

core/src/network/rpc/client.rs
```diff
@@ -105,6 +111,87 @@ impl GenesisHash {
...
+			},
+			Poll::Ready(Some(Ok(item))) => {
+				trace!("Received {first_label} item");
+				return Poll::Ready(Some(Ok(item)));
+			},
+			Poll::Ready(Some(Err(e))) => {
+				error!("Error in Avail Headers stream: {:?}", e);
+				return Poll::Ready(Some(Err(e)));
+			},
+			// Do nothing here, fall into pooling the next stream
```
aterentic-ethernal, line 172: pool -> poll

aterentic-ethernal (approved):
I guess it would be fine to merge poll_both into poll_next, and slightly reduce code duplication with polling both streams in a loop

PR #725, opened 2024-10-21
Add current ws attribute to otel

vbhattaccmu:


client/src/main.rs
```diff
@@ -317,6 +317,10 @@ async fn run(
 		),
 	];
 
+	if let Some(connected_ws) = db.get(RpcNodeKey) {
+		metric_attributes.push(("fullnode_ws".to_string(), connected_ws.host));
+	}
+
```
aterentic-ethernal, line 323: Wouldn't this just set attribute initially and never update again?

PR #745, opened 2024-11-08
Project name related updates on otel

vbhattaccmu:


core/src/telemetry/otlp.rs
```diff
@@ -192,8 +193,8 @@ fn init_counters(
 	meter: Meter,
 	origin: &Origin,
 	project_name: String,
-) -> HashMap<&'static str, Counter<u64>> {
-	[
+) -> Result<HashMap<&'static str, Counter<u64>>> {
+	let counters = [
```
aterentic-ethernal, line 197: Is there any reason why this function now return `Result`?

aterentic-ethernal (commented):
I'm not sure if doing validation on initialization of metrics is the right approach. Do we want to validate configuration, or to postpone failure on the metrics initialization? Can we enforce configuration value to conform some specification (alphanumeric character + `_`)?

core/src/telemetry/otlp.rs
aterentic-ethernal: But only possible result here is OK? There is no error branch if I didn't missed something.

momosh-ethernal: no reason to return Result, yet

core/src/telemetry/otlp.rs
```diff
@@ -268,6 +275,21 @@ pub fn initialize(project_name: String, origin: &Origin, ot_config: OtelConfig)
 	})
 }
 
+fn to_snake_case(project_name: &str) -> String {
```
momosh-ethernal, line 278: do we really need this wrapper around a std fn?

core/src/types.rs
```diff
@@ -613,3 +614,82 @@ impl From<Base64> for String {
...
+	pub fn new<S: Into<String>>(name: S) -> Self {
+		ProjectName(name.into().to_case(Case::Snake))
+	}
+}
+
+impl fmt::Display for ProjectName {
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		write!(f, "{}", self.0)
+	}
+}
```
aterentic-ethernal, line 631: We can use `use derive_more::Display` and `#[derive(Display)]` instead.

core/src/types.rs
```diff
@@ -613,3 +614,82 @@ impl From<Base64> for String {
...
+
+impl Serialize for ProjectName {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+	where
+		S: Serializer,
+	{
+		serializer.serialize_str(&self.0)
+	}
+}
```
aterentic-ethernal, line 646: I think we can also use derive_more here and just implement FromString for deserialization.

core/src/telemetry/otlp.rs
```diff
@@ -208,7 +211,11 @@ fn init_counters(
 	.iter()
 	.filter(|counter| MetricCounter::is_allowed(counter, origin))
 	.map(|counter| {
-		let otel_counter_name = format!("{project_name}.{name}", name = counter.name());
+		let otel_counter_name = format!(
+			"{project}.{name}",
+			name = counter.name(),
+			project = project_name
+		);
```
aterentic-ethernal, line 218: Can we revert this since display is implemented for `ProjectName`?

core/src/network/p2p.rs
```diff
@@ -132,9 +132,9 @@ impl FromStr for AgentVersion {
 }
 
 impl AgentVersion {
-	fn new(project_name: String, version: &str) -> Self {
+	fn new(project_name: ProjectName, version: &str) -> Self {
 		Self {
-			base_version: format!("{project_name}-{IDENTITY_AGENT_BASE}"),
+			base_version: format!("{}-{}", project_name, IDENTITY_AGENT_BASE),
```
aterentic-ethernal, line 137: Can we revert this since display is implemented for `ProjectName`?

core/src/types.rs
```diff
@@ -613,3 +614,82 @@ impl From<Base64> for String {
...
+
+impl<'de> Deserialize<'de> for ProjectName {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+	where
+		D: Deserializer<'de>,
+	{
+		let formatted_name = String::deserialize(deserializer)?.to_case(Case::Snake);
```
aterentic-ethernal, line 653: Lets just document this in project_name field of the configuration, to make constraint explicit.

aterentic-ethernal (commented):
Looks good, just minor comments that maybe can make changes more concise and clear.

aterentic-ethernal (approved):

PR #761, opened 2024-12-03
Update latest oltp crates' version and its APIs

vbhattaccmu:
In order to address oltp [dropouts](https://github.com/availproject/avail-test/actions/runs/12080729234/job/33691602730), the most reliable way is to update open telemetry to its latest crate.

### DIFF
--- a/bootstrap/src/telemetry/otlp.rs
+++ b/bootstrap/src/telemetry/otlp.rs
@@ -78,12 +78,12 @@ pub fn initialize(
 		.with_tonic()
 		.with_endpoint(&endpoint)
 		.with_protocol(Protocol::Grpc)
-		.with_timeout(Duration::from_secs(10))
+		.with_timeout(Duration::from_secs(30))
 		.build()?;
 
 	let reader = PeriodicReader::builder(exporter, Tokio)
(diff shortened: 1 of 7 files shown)


bootstrap/src/telemetry/otlp.rs
```diff
@@ -68,22 +74,26 @@ pub fn initialize(
...
-				.tonic()
-				.with_export_config(export_config),
-		)
-		.with_period(Duration::from_secs(10))
-		.with_timeout(Duration::from_secs(15))
+	let exporter = MetricExporter::builder()
+		.with_tonic()
+		.with_endpoint(&endpoint)
+		.with_protocol(Protocol::Grpc)
+		.with_timeout(Duration::from_secs(10))
```
sh3ll3x3c, line 81: Maybe we should think about increasing the default timeout to 30s?

sh3ll3x3c (approved):

bootstrap/src/telemetry/otlp.rs
vbhattaccmu: yes will increase it

aterentic-ethernal (approved):
Not sure how using grpc affects changes, but it looks better than before.

PR #764, opened 2024-12-09
feat(Tracker): Added Wallet Address For LC Validation

Mridul-Web3:


vbhattaccmu:
Hello,
Currently wallet address is stored as a key within DHT keyspaces of the Light Client, hence we wont need to track them separately through config. You can use the storage key to store any data you need.

Mridul-Web3:
@vbhattaccmu 
Thank you for the clarification regarding wallet addresses being stored as keys in the DHT. I’d appreciate some guidance to implement this effectively:

1. Could you provide an example of how to store and retrieve data (e.g., wallet addresses) using the DHT APIs?
2. Are there best practices for key naming to avoid collisions, or should we follow any existing conventions?
3. Any considerations for querying DHT data efficiently across nodes, especially for my tracker dashboard?

If there's documentation or code references for the DHT APIs in the Light Client, sharing those would be really helpful.

vbhattaccmu:
> @vbhattaccmu Thank you for the clarification regarding wallet addresses being stored as keys in the DHT. I’d appreciate some guidance to implement this effectively:
> 
> 1. Could you provide an example of how to store and retrieve data (e.g., wallet addresses) using the DHT APIs?
> 2. Are there best practices for key naming to avoid collisions, or should we follow any existing conventions?
> 3. Any considerations for querying DHT data efficiently across nodes, especially for my tracker dashboard?
> 
> If there's documentation or code references for the DHT APIs in the Light Client, sharing those would be really helpful.

currently there is an avail-address mapped to an LC and its peer id and its unique (guaranteed) for an LC. Get and Puts are automatically handled by LC itself so there is no need to manage it separately. 

When an LC is spun up, the address is generated automatically and is visible in logs. You can query the peer id by the [diagnostics api ](https://github.com/availproject/avail-light/tree/main/core/src/api/diagnostics). In case you need to query avail address in some form and use it on your end, you can raise an issue on our repository and we will look into it.

As far as a tracker of any sort is concerned, the present set of V2 and diagnostic APIs can serve all the purposes, I dont see the need to query avail-address, you can use peer id instead. please let me know if you have any questions.

PR #776, opened 2025-01-22
Remove mutable metric attributes

aterentic-ethernal:


vbhattaccmu (approved):

client/src/main.rs
```diff
@@ -297,39 +291,34 @@ async fn run(
...
-	let metric_attributes = vec![
+	let resource_attributes = vec![
 		("version", version.to_string()),
 		("role", "lightnode".to_string()),
 		("origin", cfg.origin.to_string()),
 		("peerID", peer_id.to_string()),
 		("avail_address", identity_cfg.avail_public_key),
 		("network", Network::name(&cfg.genesis_hash)),
 		("client_id", client_id.to_string()),
-		("execution_id", execution_id.to_string()),
```
moraesjeremias, line 315: We didn't observed `execution_id` changes very often

moraesjeremias: resolved on #L319

core/src/telemetry/mod.rs
```diff
@@ -2,9 +2,7 @@ pub mod otlp;
 
 use crate::types::Origin;
 
-pub const ATTRIBUTE_MULTIADDRESS: &str = "multiaddress";
 pub const ATTRIBUTE_OPERATING_MODE: &str = "operating_mode";
```
moraesjeremias, line 5: Do you think operation mode auto switch can be a source of issues in the future?

moraesjeremias (approved):
Overall LGTM,
Just minor comments/reflections.

core/src/telemetry/mod.rs
aterentic-ethernal: It can cause duplicates by name, but it is supposed to work like that, and delta would be zero for switched value. So, I don't know, but I don't see it will be big issue, and we want to redesign it anyway.

PR #779, opened 2025-01-29
Update avail rust dependency on LC

vbhattaccmu:


Cargo.toml
```diff
@@ -18,8 +18,8 @@ repository = "https://github.com/availproject/avail-light"
 
 [workspace.dependencies]
 # Internal deps
-avail-rust = { git = "https://github.com/availproject/avail.git", rev = "741aa316" }
```
aterentic-ethernal, line 21: Can we use tag or at least revision?

core/src/network/rpc/client.rs
```diff
@@ -289,12 +298,12 @@ impl<D: Database> Client<D> {
 
 	// Creates the RPC client by connecting to the provided RPC host and checks if the provided genesis hash matches the one from the client
 	async fn create_rpc_client(host: &str, expected_genesis_hash: &str) -> Result<(SDK, Node)> {
-		let client = SDK::new_insecure(host)
+		let client = SDK::new(host)
 			.await
-			.map_err(|e| Report::msg(ClientCreationError::SdkFailure(eyre!("{e}"))))?;
+			.map_err(|e| Report::msg(ClientCreationError::SdkFailure(eyre!("{:?}", e))))?;
```
aterentic-ethernal, line 303: Any reason to change this?

core/src/network/rpc/client.rs
```diff
@@ -649,12 +649,9 @@ impl<D: Database> Client<D> {
...
 				async move {
-					client
-						.rpc
-						.kate
-						.query_proof(cells.to_vec(), Some(block_hash))
+					query_proof(&client.client, cells.to_vec(), Some(block_hash))
 						.await
-						.map_err(|error| subxt::Error::Other(format!("{error}")))
+						.map_err(|error| subxt::Error::Other(format!("{:?}", error)))
 						.map_err(Into::into)
```
aterentic-ethernal, line 650: Since now query proof returns subxt error, we can replace code with, lets check other related changes also
```rust
Ok(query_proof(&client.client, cells.to_vec(), Some(block_hash)).await?) 
```

for format, we can also use 
```rust
format!("{error:?}")
```

if needed

core/src/network/rpc/client.rs
aterentic-ethernal: Error now has to_string() function:

```rust
		let client = SDK::new(host).await.map_err(|e| Report::msg(e.to_string()))?;
```

Cargo.toml
```diff
@@ -18,8 +18,8 @@ repository = "https://github.com/availproject/avail-light"
...
-dusk-plonk = { git = "https://github.com/availproject/plonk.git", tag = "v0.12.0-polygon-2" }
```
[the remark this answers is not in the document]
aterentic-ethernal: Are you sure, did you tried wasm build?

aterentic-ethernal (approved):

PR #790, opened 2025-02-18
fix/release-ubuntu-version

moraesjeremias:
# Changes
- [x] Temp removes arm64 build
- [x] Bumps Ubuntu runner version

# Reason for Change
- Ubuntu v20.0.4 runners will be sunset soon
- `openssl-sys v0.9.105` crate fails to build on arm64 due to missing runner OS openssl config. Needs to remove this build for the time being in order to unblock other builds/features. We still need to troubleshoot and fix it in the long run.


aterentic-ethernal (approved):
Don't forget to squash 🙏

PR #791, opened 2025-02-19
Set releases ubuntu version to 22

aterentic-ethernal:


.github/workflows/avail-light-bootstrap.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
```
moraesjeremias, line 16: Any concerns to bump it to 24.04?

moraesjeremias (changes requested):

.github/workflows/avail-light-bootstrap.yml
aterentic-ethernal: Fails to run on hex due to the version of glibc (requires 2.38, and available is 2.35).

moraesjeremias (approved):

PR #806, opened 2025-03-26
Download and update light client

aterentic-ethernal:


client/src/main.rs
```diff
@@ -754,6 +768,12 @@ pub async fn main() -> Result<()> {
 
 	let reason = shutdown.completed_shutdown().await;
 
+	let restart = restart.lock().await;
+	if *restart {
+		info!("Restarting Light Client...");
+		utils::restart();
+	}
```
momoshell, line 775: is that `restart` variable used elsewhere, 
if not, can we use something like this?

```suggestion
if *restart.lock().await {
    info!("Restarting Light Client...");
    utils::restart();
}
```

momoshell (approved):

core/src/updater/mod.rs
```diff
@@ -56,16 +126,68 @@ fn format_duration(duration: Duration) -> String {
...
+	let path = tempdir.path().join(name);
+
+	let tarball = File::create(&path)?;
+
+	Ok((tarball, path))
+}
+
+fn extract_tarball(tarball_path: &PathBuf, asset_name: &str) -> Result<()> {
+	debug!("File::open: {tarball_path:?}");
+	let tarball = File::open(tarball_path).unwrap();
```
sh3ll3x3c, line 140: Do we want to panic here?

sh3ll3x3c (approved):

core/src/updater/mod.rs
```diff
@@ -56,16 +126,68 @@ fn format_duration(duration: Duration) -> String {
...
+fn create_tarball(asset: &Asset, tempdir: &TempDir) -> Result<(fs::File, PathBuf)> {
+	let name = format!("{}.tar.gz", asset.target.asset_name());
+	let path = tempdir.path().join(name);
+
+	let tarball = File::create(&path)?;
+
+	Ok((tarball, path))
+}
+
+fn extract_tarball(tarball_path: &PathBuf, asset_name: &str) -> Result<()> {
```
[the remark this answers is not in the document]
aterentic-ethernal: Its not manual, problem is that we want to extract specific file from archive based on asset name, which is possible with tar command also, but I'm not sure on the tradeoffs 🤷

PR #822, opened 2025-04-25
Expose pending kbucket timeout on kademlia configuration

vbhattaccmu:


core/src/network/p2p/configuration.rs
```diff
@@ -117,6 +119,7 @@ impl Default for KademliaConfig {
 			max_kad_provided_keys: 1024,
 			operation_mode: KademliaMode::Client,
 			automatic_server_mode: true,
+			kbucket_pending_timeout: Duration::from_secs(60),
```
sh3ll3x3c, line 122: I would definitely be very careful about changing the default value here.

sh3ll3x3c (approved):

aterentic-ethernal (approved):

PR #823, opened 2025-04-25
Using Subxt RPC client without retry strategies

momoshell:
The exposed avail-rust SDK client is created with the default exponential backoff strategy, which conflicted with our strategy for retries when creating a new RPC client.

Now, we're using plain Subxt RPC client, without reconnects.


aterentic-ethernal (approved):
Just don't forget versions and changelog (depending which PR we will merge first).

PR #826, opened 2025-04-30
Created new metric for RPC host swtiches

momoshell:
Telemetry metric counters for initial and switched RPC host connections

### DIFF
--- a/core/src/api/types.rs
+++ b/core/src/api/types.rs
@@ -501,7 +501,7 @@ impl TryFrom<RpcEvent> for Option<PublishMessage> {
 				.map(Box::new)
 				.map(PublishMessage::HeaderVerified)
 				.map(Some),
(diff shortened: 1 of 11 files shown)


core/src/api/types.rs
```diff
@@ -501,7 +501,7 @@ impl TryFrom<RpcEvent> for Option<PublishMessage> {
 				.map(Box::new)
 				.map(PublishMessage::HeaderVerified)
 				.map(Some),
-			RpcEvent::ConnectedHost(_) => Ok(None), // silently skip ConnectedHost event
```
aterentic-ethernal, line 504: Do we want catch all here or just list skipped events with `EventA | EventB => Ok(None)`?

momoshell: Ok, will go with the OR match, for the case of future developments

PR #832, opened 2025-05-09
Bootstrap using the Network package from Core

momoshell:


core/src/network/p2p/configuration.rs
```diff
@@ -91,39 +91,37 @@ pub struct AutoNATConfig {
 	/// Client configuration:
 	/// Interval in which the NAT status should be re-tried if it is currently unknown or max confidence was not reached yet. (default: 90 sec)
 	#[serde(with = "duration_seconds_format")]
-	pub autonat_retry_interval: Duration,
+	pub retry_interval: Duration,
```
aterentic-ethernal, line 94: Is this breaking change, since name of the configuration parameter is changed?

aterentic-ethernal (approved):

PR #857, opened 2025-06-03
P2P client restart

sh3ll3x3c:
- implement periodic optional p2p client restart feature
- `p2p_restart_manager` is responsible for maintaining p2p client integrity during restarts:
-- graceful shutdowns of event loop and listeners
-- restarts of of listeners and event loops
-- bootstrap process

- `forward_p2p_events` used to decouple event senders and receivers in order to persist across restart sessions
- modified all p2p clients to accommodate consistent p2p client use during restarts

### DIFF
--- a/core/src/network/p2p/restart.rs
+++ b/core/src/network/p2p/restart.rs
@@ -87,7 +87,7 @@ pub async fn p2p_restart_manager(
 	mut current_shutdown: Controller<String>,
 ) {
 	// Randomize restart intervals
-	let randomized_duration = randomize_duration(restart_interval);
+	let mut randomized_duration = randomize_duration(restart_interval);
 
 	let mut interval = tokio::time::interval(randomized_duration);
 	interval.tick().await;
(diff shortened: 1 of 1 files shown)


core/src/network/p2p/restart.rs
```diff
@@ -0,0 +1,165 @@
...
+							warn!("Error stopping listeners during restart: {e}");
+						}
+					}
+				}
+
+				// Trigger shutdown of the current event loop
+				let _ = current_shutdown.trigger_shutdown("P2P client periodic restart".to_string());
+
+				// Wait a bit for graceful shutdown
+				tokio::time::sleep(Duration::from_secs(5)).await;
```
aterentic-ethernal, line 110: We can await instead:

```rust
_ = current_shutdown.completed_shutdown().await;
```

core/src/network/p2p/restart.rs
```diff
@@ -0,0 +1,165 @@
...
+	peer_id: PeerId,
+	version: String,
+	restart_interval: Duration,
+	app_shutdown: Controller<String>,
+	db: DB,
+	event_tx: UnboundedSender<OutputEvent>,
+	mut current_shutdown: Controller<String>,
+) {
+	let mut interval = tokio::time::interval(restart_interval);
+	interval.tick().await;
```
aterentic-ethernal, line 89: Maybe we need to introduce some optional randomization, for setups like ours.

PR #874, opened 2025-06-16
add diagnostics APIs for DHT metrics

vbhattaccmu:


client/src/main.rs
```diff
@@ -623,6 +624,7 @@ impl ClientState {
 			self.metrics
 				.record(MetricValue::DHTPutSuccess(success_rate));
 			self.metrics.record(MetricValue::DHTPutDuration(time_stat));
+			db.put(DhtPutSuccessKey, success_rate);
```
momoshell, line 627: We're using DB writes for every handled put request.
This will become a large overhead, just to be able to read it in the CI test.

And again, it should be under the build feature flag.

core/src/api/diagnostics/README.md
```diff
@@ -117,6 +117,67 @@ Content-Type: application/json
 }
 ```
 
+## GET `/v1/p2p/dht/{metric}`
```
momoshell, line 120: We don't want this exposed.
Just built with a feature flag for the CI tests.

core/src/api/diagnostics/mod.rs
```diff
@@ -43,11 +47,22 @@ fn p2p_peer_multiaddr_route(
...
+	warp::path!("v1" / "p2p" / "dht" / String)
+		.and(warp::get())
+		.and(warp::any().map(move || db.clone()))
+		.then(get_dht_metric)
+		.map(log_internal_server_error)
+}
+
 #[allow(clippy::too_many_arguments)]
-pub fn routes(p2p_client: Client) -> P2PFilter {
+pub fn routes(p2p_client: Client, db: impl Database + Clone + Send + Sync + 'static) -> P2PFilter {
```
momoshell, line 61: Same here, needs to be under a feature flag.

core/src/api/server.rs
```diff
@@ -74,7 +74,7 @@ impl<T: Database + Clone + Send + Sync + 'static> Server<T> {
 
 		// Add P2P routes depending on whether P2P is enabled
 		let p2p_routes = if let Some(p2p_client) = &self.p2p_client {
-			diagnostics::routes(p2p_client.clone())
+			diagnostics::routes(p2p_client.clone(), self.db.clone())
```
momoshell, line 77: A feature flag for CI tests is needed.

momoshell:
We need to revert these changes ASAP
@aterentic-ethernal @sh3ll3x3c @jakubcech

PR #958, opened 2026-05-13
fix: check gcell block constraints at proof verify step

vbhattaccmu:


aterentic-ethernal (approved):
The only concern is one malicious peer can invalidate entire batch, but I guess that was already a case before this PR.

PR #960, opened 2026-07-31
fix(client): preserve config values and apply AutoNAT CLI override

grandpig:
## Summary

Fix CLI configuration merging so that omitted command-line options no longer overwrite values loaded from the configuration file.

This change:

- makes `--auto-nat-mode` optional and applies it only when explicitly provided;
- preserves `no_update` and `tracking_service_enable` when they are enabled in the config file;
- preserves `operator_address` when the corresponding CLI option is omitted;
- adds regression tests covering the configuration file and CLI merge behavior.

## Problem

`--auto-nat-mode` was defined with `disabled` as its CLI default, but its value was never applied to `cfg.libp2p.behaviour.auto_nat_mode`. As a result, passing `--auto-nat-mode enabled` had no effect.

Using a non-optional default would also make it impossible to distinguish an omitted CLI option from an explicit `disabled` value, causing configuration-file values to be overwritten.

Additionally, the following assignments always replaced values loaded from the configuration file with CLI defaults:

```rust
cfg.no_update = opts.no_update;
cfg.tracking_service_enable = opts.tracking_service_enable;
cfg.operator_address = opts.operator_address.clone();
```

When those options were omitted, true values were reset to false and an existing operator address was reset to None.

## Fix

Change auto_nat_mode to Option<AutoNatMode> and apply it only when supplied.

Use additive behavior for the enable-only boolean flags, and update the operator address only when the CLI option is present.

I have tested by the following cmd:

```shell
cargo test -p avail-light-client --bin avail-light-client
cargo test -p avail-light-client --no-default-features --bin avail-light-client
cargo clippy -p avail-light-client --no-default-features -- -D warnings
```


aterentic-ethernal (approved):

aterentic-ethernal:
Closing PR to retrigger failing workflow by reopening it, due to missing token for test dependency fetching.

PR #961, opened 2026-08-03
ci: fall back to GITHUB_TOKEN for grcov install on fork PRs

aterentic-ethernal:
The `cargo test` job fails on every pull request opened from a fork, before the coverage report is generated. Tests themselves pass; the job dies at the **Install grcov** step with `Input required and not supplied: github-token`. This makes the fix a prerequisite for merging any external contribution.

The cause is that GitHub withholds repository secrets from workflows triggered by `pull_request` events originating in a fork. `secrets.PAT_TOKEN` therefore expands to an empty string, and because Actions drops empty inputs entirely, `SierraSoftworks/setup-grcov` sees its required `github-token` input as absent rather than blank.

## How it works

- `secrets.PAT_TOKEN || secrets.GITHUB_TOKEN` keeps the PAT on branch pushes and same-repo pull requests, where it is available and carries higher API rate limits.
- On fork pull requests the expression falls through to `GITHUB_TOKEN`, which is minted per job rather than stored as a repository secret and so remains available.
- `GITHUB_TOKEN` is forced read-only for fork pull requests. Downloading a grcov release asset needs only read access, so the reduced scope is sufficient.
- No additional credential is exposed. `GITHUB_TOKEN` is already present in every job context; passing it as an action input does not widen what a workflow can reach.

## Caveats

- Withholding `PAT_TOKEN` from fork pull requests is a deliberate protection, not a misconfiguration. Routing around it with `pull_request_target` would run the base workflow with full secrets and is deliberately not done here.
- `arduino/setup-protoc` reads the same secret via `repo-token` and is left unchanged. It treats the token as optional, so on fork pull requests it silently falls back to unauthenticated API calls and could hit rate limits under load. Worth a follow-up, but it is not currently failing.
- The `codecov/codecov-action` upload step runs without a token on fork pull requests. Tokenless upload generally works for public repositories, so it is untouched here.
- Open pull requests pick this up once their merge ref is recomputed against `main`; no rebase is needed. Re-running an existing failed run reuses the old commit and will not see the change, so a fresh run is required.


brixial (approved):
LGTM

